PR #201, opened 2018-07-26
Bugfix/uncommitted balance

martineehrlich:
## Description
Uncommitted balance was being calculated incorrectly because we were mistakenly using the actual uncommitted balance from lnd engine as the total balance. The fix is to not subtract the committed balance from the uncommitted balance, and instead just leave the values as is. 

## Related PRs

## Todos
- [x] Tests
- [x] Documentation
- [x] Link to Trello

### DIFF
--- a/broker-daemon/broker-rpc/wallet-service/get-balances.js
+++ b/broker-daemon/broker-rpc/wallet-service/get-balances.js
@@ -7,14 +7,14 @@
  * @return {Object} res.engine
  */
 async function getEngineBalances ([symbol, engine]) {
-  const [totalBalance, totalChannelBalance] = await Promise.all([
+  const [uncommittedBalance, totalChannelBalance] = await Promise.all([
     engine.getTotalBalance(),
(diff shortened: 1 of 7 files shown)


martineehrlich:
[Fix negative uncommitted balance in wallet balance](https://trello.com/c/e8sZhU0l/534-fix-negative-uncommitted-balance-in-wallet-balance)

broker-daemon/broker-rpc/wallet-service/get-balances.js
```diff
@@ -7,14 +7,14 @@
  * @return {Object} res.engine
  */
 async function getEngineBalances ([symbol, engine]) {
-  const [totalBalance, totalChannelBalance] = await Promise.all([
+  const [uncommittedBalance, totalChannelBalance] = await Promise.all([
     engine.getTotalBalance(),
```
treygriffith: should rename this on the engine as well

PR #207, opened 2018-07-28
Bugfix: use a named volume to persist sparkswapd data between restarts

treygriffith:
## Description
We had an issue where we were losing block orders when we `down`ed containers on `sparkswapd`. This is because we weren't using a named volume, so docker compose was creating a new volume when the service came back `up`. Notably, this wasn't a problem when recreating containers, only when explicitly `down`ing them.

## Related PRs

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Link to Trello

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -14,7 +14,8 @@ services:
       - './proto:/home/app/proto/'
       - '/home/app/node_modules'
       - './node_modules/lnd-engine:/home/app/node_modules/lnd-engine'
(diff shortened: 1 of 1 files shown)


treygriffith:
[BUG: all block orders deleted when KBD is restarted](https://trello.com/c/FJOkAJt7/329-bug-all-block-orders-deleted-when-kbd-is-restarted)

PR #209, opened 2018-07-29
Feature - Add value to commit-balance

dannypaz:
## Description
This PR adds an extra option for `<amount>` when using commit-balance. Currently we only allow the user to commit the full balance of their wallet, however at the current time we do not account for FEEs.  If the user does not have an uncommitted balance greater than the maximum channel balance + fees, then the command will fail.

There is a follow-up ticket to take fees into account.

Additionally, there are some docker-compose changes that were needed during testnet testing (typos that never made it over to master branch)

## Related PRs

## Todos
- [x] Tests
- [x] Documentation
- [x] Link to Trello


dannypaz:
[Add amount to CLI for broker during `commit-balance`](https://trello.com/c/9wDLxhiR/543-add-amount-to-cli-for-broker-during-commit-balance)

broker-cli/commands/wallet.js
```diff
@@ -130,29 +130,42 @@ async function commitBalance (args, opts, logger) {
...
       return logger.info('Your current uncommitted balance is 0, please add funds to your daemon')
     }
-    // We try to take the `Math.min` total here between 2 Big numbers due to a
+
+    // We try to take the lowest total here between 2 Big numbers due to a
     // commit limit specified as MAX_CHANNEL_BALANCE
     let maxSupportedBalance = uncommittedBalance
 
     if (totalUncommittedBalance.gt(ENUMS.MAX_CHANNEL_BALANCE)) {
       maxSupportedBalance = ENUMS.MAX_CHANNEL_BALANCE
```
treygriffith: shouldn't this be per currency?

broker-cli/commands/wallet.js
```diff
@@ -130,29 +130,42 @@ async function commitBalance (args, opts, logger) {
...
     if (!ACCEPTED_ANSWERS.includes(answer.toLowerCase())) return
 
+    if (Big(maxSupportedBalance).gte(uncommittedBalance)) {
+      throw new Error(`Amount specified is larger than your current uncommitted balance of ${Big(uncommittedBalance).div(divideBy)} ${symbol}`)
+    }
+
     await client.walletService.commitBalance({ balance: maxSupportedBalance.toString(), symbol })
 
-    logger.info('Successfully committed balance to SparkSwap Relayer!')
+    logger.info('Successfully committed balance to sparkswap Relayer!')
```
treygriffith: this is definitely not for this ticket, but I think we need some way to check on the status of a commitment to the Relayer. Particularly since we've seen most of the problems crop up when one out of the two channels are open - that's a confusing state for a user unless they know A LOT about the inner workings.

treygriffith: (I opened a ticket: https://trello.com/c/W4XTd5wl/555-better-user-feedback-on-balance-commitment-status)

broker-cli/commands/wallet.js
```diff
@@ -170,7 +183,7 @@ module.exports = (program) => {
 
       // TODO: Figure out a way to handle subArguments that could be dynamic
       // for each command
-      let [symbol = ''] = subArguments
+      let [symbol = '', amount] = subArguments
```
treygriffith: why does `symbol` default to empty string but `amount` doesn't?

dannypaz: @treygriffith we can default it for now

broker-cli/commands/wallet.js
dannypaz: @treygriffith I added a TODO and have thrown an issue in our backlog.

PR #210, opened 2018-07-30
FIX docker-compose to prod

dannypaz:
## Description
Updates to docker-compose to fix missing volume and simnet relayer urls

## Related PRs

## Todos
- [n/a] Tests
- [n/a] Documentation
- [n/a] Link to Trello

### DIFF
--- a/docker-compose.prod.yml
+++ b/docker-compose.prod.yml
@@ -54,7 +54,7 @@ services:
     environment:
       - RPC_USER=kuser
       - RPC_PASS=kpass
-      - RPC_HOST=docker.for.mac.host.internal:18556
+      - RPC_HOST=btcd:18556
       - RPC_CERT_PATH=/shared/rpc-btc.cert
       - ADMIN_MACAROON=/shared/lnd-engine-admin-btc.macaroon
       - READ_ONLY_MACAROON=/shared/lnd-engine-readonly-btc.macaroon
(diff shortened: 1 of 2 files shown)


docker-compose.prod.yml
```diff
@@ -32,9 +32,9 @@ services:
       # This address is specifically tied to the relayer network
       # This is a temporary workaround for preventing the use of external networks
       # for the relayer. In a testnet/mainnet setting, this will be a public IP.
```
treygriffith: this comment is out of date now, right?

docker-compose.prod.yml
```diff
@@ -57,7 +54,7 @@ services:
     environment:
       - RPC_USER=kuser
       - RPC_PASS=kpass
-      - RPC_HOST=btcd:18556
+      - RPC_HOST=docker.for.mac.host.internal:18556
```
treygriffith: is it really docker.for.mac when deployed? And why are we not using `btcd` as the host name?

docker-compose.prod.yml
```diff
@@ -71,7 +68,7 @@ services:
       - LISTEN=0.0.0.0:10113
       # This address is for development only. You must change this to a
       # tld to use in a production setting
-      - EXTERNAL_ADDRESS=your.daemon.address:10113
+      - EXTERNAL_ADDRESS=docker.for.mac.host.internal:10113
```
treygriffith: this seems wrong for a hosted version

PR #212, opened 2018-08-02
Feature - ENV commands for multiple environments

dannypaz:
## Description
This PR adds the following commands to `npm` to allow the user to start up the broker on different environments w/ out having to specify docker-compose files explicitly

```
npm run simnet
npm run testnet
npm run mainnet
```

## Related PRs

## Todos
- [n/a] Tests
- [ ] Documentation
- [x] Link to Trello


dannypaz:
[Add command for upping broker through bin](https://trello.com/c/mPl7cMQh/590-add-command-for-upping-broker-through-bin)

scripts/build.sh
```diff
@@ -29,6 +29,7 @@ if [ -d ./proto ]; then
 fi
 
 git clone https://github.com/sparkswap/relayer-proto.git ./proto
+rm -rf ./proto/.git
```
treygriffith: why did we make this a submodule? and why in this pr?

scripts/simnet.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+
+##################################################################
+# Starts the broker docker stack on simnet
+##################################################################
+
+set -e -u
+
+docker-compose up -d
```
treygriffith: don't you still need prod?

scripts/build.sh
dannypaz: @treygriffith this change was made in a prior PR and git only gives you a warning for this change. I can make the change in another PR or we can leave it here

scripts/simnet.sh
dannypaz: @treygriffith technically there are 2 versions of simnet. Hosted and Local. We have changes coming that will change the build process (using simnet instead of local dev) This is probably OK for now.

PR #216, opened 2018-08-03
FIX - remove local funding from broker

dannypaz:
## Description
`npm run fund` and the fund-simnet-local files are sparkswap specific because the relayer is not currently OSS. This PR removes these commands for clarity in the broker

## Related PRs

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Link to Trello


treygriffith:
Do we have an alternative way to do this?

dannypaz:
@treygriffith this is resolved in https://github.com/sparkswap/relayer/pull/134

PR #218, opened 2018-08-03
Fix: better handling for sparkswapd failing to start

treygriffith:
## Description
This change shuts down the sparkswapd process if it fails to start so that pm2 can restart it.

It also improves logging from these failures to make it more clear what the issue is - particularly from the most common issue of lnd not being ready.

## Related PRs

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Link to Trello

### DIFF
--- a/broker-daemon/utils/logger.js
+++ b/broker-daemon/utils/logger.js
@@ -28,6 +28,7 @@ if (process.env.NODE_ENV !== 'production') {
   logger.add(new winston.transports.Console({
     format: winston.format.combine(
       winston.format.timestamp(),
+      winston.format.colorize(),
       winston.format.simple()
     )
   }))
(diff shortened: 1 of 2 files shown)


broker-daemon/utils/logger.js
```diff
@@ -28,6 +28,7 @@ if (process.env.NODE_ENV !== 'production') {
   logger.add(new winston.transports.Console({
     format: winston.format.combine(
       winston.format.timestamp(),
+      winston.format.colorize(),
```
dannypaz: just as a note, this makes it harder to parse the output of logs (it adds extra characters to the output)

dannypaz:
@treygriffith PM2 will only restart the process 10 times on the broker https://github.com/sparkswap/broker/blob/master/pm2.json#L5 before the container is set to an error status (and must be restarted through `docker-compose`.

The restarting happens pretty quickly, so the edge case of lnd not being up because of blockchain indexing, still exists.

Unfortunately I dont have any suggestions for a solution on that yet, but this PR is a step in the right direction.

PR #219, opened 2018-08-03
add pending wallet balances to committed and uncommitted sections

martineehrlich:
## Description
The wallet balance command on the broker needs pending balances because in the process of commit a balance to a channel or closing a channel, the funds go into a pending state. Without this change, the funds completely disappear and this is confusing and worrisome for the end user. 

1) Fix uncommitted balance to be the correct value from the LND api
2) Add a pending channel balance which is the amount of funds in the process of being committed.
3) Add a pending uncommitted channel balance which is the amount of funds in channels that are being closed combined with the amount of unconfirmed funds in the wallet. 

## Related PRs

## Todos
- [x] Tests
- [x] Documentation
- [x] Link to Trello


broker-cli/proto/broker.proto
```diff
@@ -791,12 +796,16 @@ service WalletService {
...
-   *     "totalChannelBalance": "16700000"
+   *     "totalChannelBalance": "16700000",
+   *     "totalPendingChannelBalance": "10000",
+   *     "uncommittedPendingBalance": "00000"
    *   },
    *   {
    *     "symbol": "LTC",
    *     "uncommittedBalance": "1000000000",
-   *     "totalChannelBalance": "16700000"
+   *     "totalChannelBalance": "16700000",
```
treygriffith: you should update the shell example

dannypaz:
[wallet balance should reflect pending actions](https://trello.com/c/SZdNqKUV/602-wallet-balance-should-reflect-pending-actions)

PR #229, opened 2018-08-09
Feature: remove the existing orderbook when receiving a signal from the relayer

treygriffith:
## Description
This change allows the broker to reset its orderbook when indicated by the Relayer. This avoids problems where old events get stuck in the broker's orderbook with no way to remove them.

The Relayer's associated change should be deployed before this one, since it depends on changes to the Relayer proto definition.

## Todos
- [x] Tests
- [ ] Documentation
- [ ] Link to Trello

### DIFF
--- a/broker-daemon/relayer/relayer-client.js
+++ b/broker-daemon/relayer/relayer-client.js
@@ -6,7 +6,7 @@ const caller = require('grpc-caller')
 const Identity = require('./identity')
 
 const { MarketEvent } = require('../models')
-const { loadProto } = require('../utils')
+const { loadProto, migrateStore } = require('../utils')
 
 /**
  * @constant
@@ -97,6 +97,8 @@ class RelayerClient {
 
       try {
         const watcher = this.orderbookService.watchMarket(params)
+        // this flag tells us if we are the in the middle of migrating the db
+        let migrating = false
 
         watcher.on('end', () => {
           this.logger.info('Remote ended stream', params)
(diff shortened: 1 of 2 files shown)


broker-daemon/relayer/relayer-client.js
```diff
@@ -97,6 +97,8 @@ class RelayerClient {
 
       try {
         const watcher = this.orderbookService.watchMarket(params)
+        // this flag tells us if we are the in the middle of migrating the db
+        let migrating = false
```
dannypaz: [nit] use `null` or let the value be undefined. It might be weird to set this as a boolean and then assign it to a function later in the method

broker-daemon/relayer/relayer-client.js
```diff
@@ -97,6 +97,8 @@ class RelayerClient {
 
       try {
         const watcher = this.orderbookService.watchMarket(params)
+        // this flag tells us if we are the in the middle of migrating the db
```
dannypaz: Add the `why` to this just to be direct about why we need the flag.

broker-daemon/relayer/relayer-client.spec.js
```diff
@@ -3,7 +3,7 @@ const { sinon, rewire, delay, expect } = require('test/test-helper')
 
 const RelayerClient = rewire(path.resolve('broker-daemon', 'relayer', 'relayer-client'))
 
-describe('RelayerClient', () => {
+describe.only('RelayerClient', () => {
```
dannypaz: remove only

broker-daemon/relayer/relayer-client.spec.js
```diff
@@ -280,6 +286,54 @@ describe('RelayerClient', () => {
...
+
+      expect(migrateStore).to.have.been.calledOnce()
+      expect(migrateStore).to.have.been.calledWith(store, store, sinon.match.func)
+
+      const migrator = migrateStore.args[0][2]
+
+      expect(migrator('hello')).to.be.eql({ type: 'del', key: 'hello' })
+    })
+
+    it('waits for migrating to be done before processing more events', async () => {
```
dannypaz: +1 for this test

dannypaz:
@treygriffith these changes LGTM.

PR #235, opened 2018-08-15
Smart commit

martineehrlich:
## Description
We want to check preexisting channel balances on commit so that we don't unnecessarily open channels if they already exist. If channels are already open during commit, check balances, if the balance is low, ask user to release channels, if balance is enough do not open channels.

## Related PRs
https://github.com/sparkswap/lnd-engine/pull/91

## Todos
- [x] Tests
- [x] Documentation
- [x] Link to Trello

### DIFF
--- a/broker-daemon/broker-rpc/wallet-service/commit-balance.js
+++ b/broker-daemon/broker-rpc/wallet-service/commit-balance.js
@@ -1,5 +1,5 @@
 const { PublicError } = require('grpc-methods')
-const { convertBalance } = require('../../utils')
+const { convertBalance, Big } = require('../../utils')
 const { currencies } = require('../../config')
 /**
  * @constant
@@ -71,9 +71,19 @@ async function commitBalance ({ params, relayer, logger, engines }, { EmptyRespo
     throw new PublicError(`Maxium balance of ${MAX_CHANNEL_BALANCE} exceeded for committing to the relayer. Please try again.`)
   }
 
+  const {maxBalance: maxOutboundBalance} = await engine.getMaxChannel()
+  const {maxBalance: maxInboundBalance} = await inverseEngine.getMaxChannel({outbound: false})
+  const convertedBalance = convertBalance(balance, symbol, inverseSymbol)
+  if (maxOutboundBalance || maxInboundBalance) {
+    if ((maxOutboundBalance && Big(maxOutboundBalance).lte(balance)) || (maxInboundBalance && Big(maxInboundBalance).lte(convertedBalance))) {
+      throw new PublicError('You have another channel with a balance lower than what you want to open, release that channel and try again.')
+    } else {
+      throw new PublicError('You already have a channel with that balance or greater.')
+    }
+  }
+
   await engine.createChannel(address, balance)
 
-  const convertedBalance = convertBalance(balance, symbol, inverseSymbol)
   const paymentChannelNetworkAddress = await inverseEngine.getPaymentChannelNetworkAddress()
 
   await relayer.paymentChannelNetworkService.createChannel({address: paymentChannelNetworkAddress, balance: convertedBalance, symbol: inverseSymbol})
--- a/broker-daemon/broker-rpc/wallet-service/commit-balance.js
+++ b/broker-daemon/broker-rpc/wallet-service/commit-balance.js
@@ -71,14 +71,31 @@ async function commitBalance ({ params, relayer, logger, engines }, { EmptyRespo
     throw new PublicError(`Maxium balance of ${MAX_CHANNEL_BALANCE} exceeded for committing to the relayer. Please try again.`)
   }
 
+  // Get the max balance for outbound and inbound channels to see if there are already channels with the balance open. If this is the
+  // case we do not need to go to the trouble of opening new channels
   const {maxBalance: maxOutboundBalance} = await engine.getMaxChannel()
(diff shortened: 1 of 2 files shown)


broker-daemon/broker-rpc/wallet-service/commit-balance.js
```diff
@@ -71,9 +71,19 @@ async function commitBalance ({ params, relayer, logger, engines }, { EmptyRespo
     throw new PublicError(`Maxium balance of ${MAX_CHANNEL_BALANCE} exceeded for committing to the relayer. Please try again.`)
   }
 
+  const {maxBalance: maxOutboundBalance} = await engine.getMaxChannel()
+  const {maxBalance: maxInboundBalance} = await inverseEngine.getMaxChannel({outbound: false})
+  const convertedBalance = convertBalance(balance, symbol, inverseSymbol)
+  if (maxOutboundBalance || maxInboundBalance) {
+    if ((maxOutboundBalance && Big(maxOutboundBalance).lte(balance)) || (maxInboundBalance && Big(maxInboundBalance).lte(convertedBalance))) {
```
dannypaz: Do we want `lte` here? I figured we would just want to throw if the balance is less than the amount

broker-daemon/broker-rpc/wallet-service/commit-balance.js
```diff
@@ -71,9 +71,19 @@ async function commitBalance ({ params, relayer, logger, engines }, { EmptyRespo
     throw new PublicError(`Maxium balance of ${MAX_CHANNEL_BALANCE} exceeded for committing to the relayer. Please try again.`)
   }
 
+  const {maxBalance: maxOutboundBalance} = await engine.getMaxChannel()
+  const {maxBalance: maxInboundBalance} = await inverseEngine.getMaxChannel({outbound: false})
+  const convertedBalance = convertBalance(balance, symbol, inverseSymbol)
+  if (maxOutboundBalance || maxInboundBalance) {
```
dannypaz: this is unneeded because we already check if these exist in the if statement below

broker-daemon/broker-rpc/wallet-service/commit-balance.js
dannypaz: I would separate this if statement so we can use `debug` logs to provide more information. Example:

```
// If the outbound balance isn't sufficient
if (!maxOutboundBalance) throw new PublicError('no channels exist or something')
if (Big(maxOutboundBalance).lt(balance)) {
  this.logger.debug('The outbound balance was less than the balance specified or something', { maxOutboundBalance, balance })
  throw new PublicError('.........')
}
etc...
```

broker-daemon/broker-rpc/wallet-service/commit-balance.js
martineehrlich: This is needed, because if neither of them exist then we open channels, but the conditions below determine what error we need to throw if either channel is open. But I guess if we break it up like you said then this is not necessary.

broker-daemon/broker-rpc/wallet-service/commit-balance.js
martineehrlich: We wouldn't throw an error if the channels did not exist, but I get the gist.

martineehrlich:
["Smart" commit](https://trello.com/c/9WwCPFJP/604-smart-commit)

broker-daemon/broker-rpc/wallet-service/commit-balance.js
```diff
@@ -71,9 +71,33 @@ async function commitBalance ({ params, relayer, logger, engines }, { EmptyRespo
...
+
+    let errorMessage
+    if (insufficientOutboundBalance) {
+      errorMessage = 'You have another outbound channel open with a balance lower than desired, release that channel and try again.'
+    } else if (insufficientInboundBalance) {
+      errorMessage = 'You have another inbound channel open with a balance lower than desired, release that channel and try again.'
+    } else {
+      errorMessage = 'You already have a channel open with that balance or greater.'
+    }
+    logger.error(`${errorMessage}, balance: ${balance}, maxOutboundBalance: ${maxOutboundBalance}, maxInboundBalance: ${maxInboundBalance}, inboundBalance: ${convertedBalance}`)
```
dannypaz: [nit] add the balances in the 2nd argument of error. It should be close to the same formatting `logger.error(errorMessage, { balance, maxOutboundBalance....})`

broker-daemon/broker-rpc/wallet-service/commit-balance.js
```diff
@@ -71,9 +71,33 @@ async function commitBalance ({ params, relayer, logger, engines }, { EmptyRespo
...
+    const insufficientOutboundBalance = maxOutboundBalance && Big(maxOutboundBalance).lt(balance)
+    const insufficientInboundBalance = maxInboundBalance && Big(maxInboundBalance).lt(convertedBalance)
+
+    let errorMessage
+    if (insufficientOutboundBalance) {
+      errorMessage = 'You have another outbound channel open with a balance lower than desired, release that channel and try again.'
+    } else if (insufficientInboundBalance) {
+      errorMessage = 'You have another inbound channel open with a balance lower than desired, release that channel and try again.'
+    } else {
+      errorMessage = 'You already have a channel open with that balance or greater.'
```
dannypaz: [nit] Use the balance that was inputted `You already have a channel open with ${balance} or greater`

dannypaz:
@martineehrlich LGTM

PR #237, opened 2018-08-15
FEATURE - Add SSL to Broker

dannypaz:
## Description
This PR adds SSL auth from the CLI to the Broker through the use of certs generated by the broker daemon.

Changes:
1. generate certs in the sparkswapd Dockerfile
2. secure broker's rpc server with ssl
3. add DISABLE_AUTH env variable to broker daemon
4. add `disableSsl` to cli configuration
5. added `rpcCert` to cli configuration
6. updated CLI to use correct GRPC cipher suite
7. updated DISABLE_AUTH to DISABLE_RELAYER_AUTH for message signing authentication for the relayer
8. Added `EXTERNAL_ADDRESS` option for `npm run build`

## Related PRs
List related PRs if applicable

## Todos
- [x] Tests
- [x] Documentation
- [x] Link to Trello

PR #238, opened 2018-08-16
Feature - Basic Auth to broker daemon

dannypaz:
## Description
This PR adds basic auth to all broker rpc calls. These changes are a part of a larger initiative to allow the user to host a daemon AND secure the connection

Basic auth is handled by a `DISABLE_AUTH` flag that is present on both the broker and CLI. This flag is set to `true` by default until a later period.

Authentication between a remote broker daemon and local CLI by passing credentials through an SSL connection in grpc.

Below are the steps required to setup a daemon for basic auth:
1. In your `~/.sparkswap.js` configuration file for the CLI, set `disableAuth: false`
2. On your broker daemon, set the env `DISABLE_AUTH=false`
3. During installation of the broker, change the line `npm run build` to `EXTERNAL_ADDRESS=your.ip.address npm run build` to allow the daemon to generate the correct rpc certs
4. On your local machine, navigate to the broker directory. ex: `cd broker`
5. copy the cert from your daemon to the host's home directory with 
`ssh <your user>@<your.ip.address> "docker cp app_sparkswapd_1:/secure/broker-rpc-tls.cert ."`
6. copy the cert from the host's home directory to your local broker-cli installation with `scp root@feeny.testnet.sparkswap.com:~/broker-rpc-tls.cert ./broker-cli/certs/feeny-rpc-tls.cert`

These steps are ONLY for testing. To make auth secure, make sure to change the `RPC_USER/RPC_PASS` credentials for both the cli and broker.

Authentication Error for Unary calls
screen shot 2018-08-16 at 2 22 30 pm

Authentication Error for Streaming Calls
screen shot 2018-08-16 at 2 22 22 pm

## Related PRs
https://github.com/sparkswap/broker/pull/237

## Todos
- [x] Tests
- [ ] Documentation
- [x] Link to Trello


dannypaz:
[Add Basic auth between KCLI and KBD](https://trello.com/c/kcxsYPLg/13-add-basic-auth-between-kcli-and-kbd)

broker-daemon/bin/sparkswapd.spec.js
```diff
@@ -77,8 +90,6 @@ describe('sparkswapd', () => {
     expect(BrokerDaemon).to.have.been.calledWith(sinon.match({ rpcAddress }))
   })
 
-  xit('defaults the rpc address to the env variable')
```
dannypaz: I tried to retest all of these default env vars again, but was unable to get it to work. I moved the comment up to the top and removed these pending tests

broker-cli/utils/basic-auth.js
```diff
@@ -0,0 +1,48 @@
...
+ * @default
+ */
+const BASIC_AUTH_PREFIX = 'Basic'
+
+/**
+ * @constant
+ * @type {String}
+ * @default
+ */
+const BASIC_AUTH_DELIMITER = ':'
```
martineehrlich: Same comment as above

broker-cli/utils/basic-auth.js
```diff
@@ -0,0 +1,48 @@
+const grpc = require('grpc')
+
+/**
+ * @constant
+ * @type {String}
+ * @default
+ */
+const BASIC_AUTH_PREFIX = 'Basic'
```
martineehrlich: Same comment as above

broker-daemon/utils/basic-auth.js
```diff
@@ -0,0 +1,46 @@
...
+
+  const rawCredentials = Buffer.from(base64Token, 'base64').toString()
+  const [username, password] = rawCredentials.split(BASIC_AUTH_DELIMITER)
+
+  logger.debug('Checking if credentials are valid between cli and broker')
+
+  // rpcUser and rpcPass are bound to context where #verify is used due to the way
+  // grpc-method handles the authorization middleware
+  if (username === this.rpcUser && password === this.rpcPass) {
+    return true
```
martineehrlich: verify should either return true/false or void/error

broker-cli/broker-daemon-client/config.js
```diff
@@ -12,9 +12,17 @@ const MISSING_FILE_MESSAGE = 'Cannot find module'
...
+ * @default
  */
 const DEFAULT_CONFIG = require('../.sparkswap.default.js')
 
+/**
+ * @type {String}
+ * @constant
+ * @default
+ */
+const USER_CONFIG_FILE = '.sparkswap.js'
```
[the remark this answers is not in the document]
dannypaz: I will change to `USER_CONFIG_FILE_PATH`. As a best practice, we should be constantizing any magic strings that are used internally for configuration (like this filepath, or project root, etc).

broker-cli/broker-daemon-client/index.js
```diff
@@ -4,29 +4,32 @@ const caller = require('grpc-caller')
...
+} = require('../utils')
 
 /**
+ * Root path from the current module used to resolve cert file paths
  * @constant
  * @type {String}
  * @default
  */
-const PROTO_PATH = path.resolve(__dirname, '..', 'proto', 'broker.proto')
+const PROJECT_ROOT = '..'
```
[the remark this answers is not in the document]
dannypaz: We probably dont need this one

broker-cli/utils/basic-auth.js
dannypaz: see my comment above https://github.com/sparkswap/broker/pull/238#discussion_r213758212

broker-cli/utils/basic-auth.js
dannypaz: see my comment above https://github.com/sparkswap/broker/pull/238#discussion_r213758212

broker-daemon/utils/basic-auth.js
dannypaz: we'll have to go with `void/error` because this method is called from inside of grpc-methods. If we returned true/false we wouldn't be able to halt the operation if auth failed.

broker-cli/broker-daemon-client/index.js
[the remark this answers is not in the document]
dannypaz: On second thought, we should leave it. PROJECT_ROOT is used twice in the file, provides a decent description of what that path is relative to the file, and if the file path every changes, we only need to edit in one place

broker-daemon/utils/basic-auth.spec.js
```diff
@@ -0,0 +1,76 @@
+const path = require('path')
+const { expect, sinon, rewire } = require('test/test-helper')
+const { PublicError } = require('grpc-methods')
+
+const basicAuth = rewire(path.resolve(__dirname, 'basic-auth'))
+const credentialGenerator = rewire(path.resolve(__dirname, '..', '..', 'broker-cli', 'utils', 'basic-auth'))
+
+describe.only('basicAuth', () => {
```
martineehrlich: Don't think this .only should be here

martineehrlich:
This looks good, except that one comment where you left .only in the test.

broker-cli/utils/basic-auth.js
```diff
@@ -0,0 +1,48 @@
...
+ * Creates a Basic Authentication string (Base64) from user/password credentials
+ *
+ * Example Result: `Basic
+ *
+ * @param {String} username
+ * @param {String} password
+ * @return {String} res - Basic Authentication string (RFC 7235)
+ */
+function credentialsToBasicAuth (username, password) {
+  const credentialHash = Buffer.from(`${username}${BASIC_AUTH_DELIMITER}${password}`).toString('base64')
```
treygriffith: I don't like calling this a hash since it's just a base64 encoding.

broker-cli/utils/basic-auth.spec.js
```diff
@@ -0,0 +1,78 @@
...
+
+    beforeEach(() => {
+      user = 'sparkswap'
+      pass = 'password'
+      credentialsToBasicAuth = basicAuth.__get__('credentialsToBasicAuth')
+    })
+
+    it('generates a basic authentication header for a username and password', () => {
+      const res = credentialsToBasicAuth(user, pass)
+      // Format of the has is based off of RFC 7235
```
treygriffith: Doesn't grpc have built-in username/pass? or do people recreate their own authentication headers like this?

broker-daemon/utils/basic-auth.js
```diff
@@ -0,0 +1,44 @@
...
+
+  // rpcUser and rpcPass are bound to context where #verify is used due to the way
+  // grpc-method handles the authorization middleware
+  if (username !== this.rpcUser || password !== this.rpcPass) {
+    logger.debug('Basic Authentication has failed. Username/Password did not match')
+    throw new PublicError('Basic Authentication Failed, please check the clis user/pass credentials')
+  }
+}
+
+module.exports = { verify }
```
treygriffith: why not export the function directly?

docker-compose.yml
```diff
@@ -46,9 +46,11 @@ services:
       - RELAYER_CERT_HOST=${RELAYER_CERT_HOST}
       - RELAYER_CERT_PATH=/secure/relayer-root.pem
       # WARNING: Only disable broker SSL during development
-      - DISABLE_AUTH=false
+      - DISABLE_AUTH=true
```
treygriffith: DISABLE_AUTH used to control authentication to the Relayer, right? Has that been changed to a different environment variable? And why is this disabled by default?

broker-daemon/broker-rpc/broker-rpc-server.js
```diff
@@ -79,21 +52,24 @@ class BrokerRPCServer {
...
+    this.rpcUser = rpcUser
+    this.rpcPass = rpcPass
+    this.auth = basicAuth.verify.bind(this)
 
     this.protoPath = path.resolve(BROKER_PROTO_PATH)
 
     this.server = new grpc.Server()
 
-    this.adminService = new AdminService(this.protoPath, { logger, relayer, engines })
+    this.adminService = new AdminService(this.protoPath, { logger, relayer, engines, basicAuth: this.auth })
```
treygriffith: I don't think the property should be named `basicAuth`, it's leaking an implementation detail into places it doesn't need to be. Any reason it can't be just `auth` ?

broker-daemon/broker-rpc/broker-rpc-server.js
```diff
@@ -79,21 +52,24 @@ class BrokerRPCServer {
     this.pubKeyPath = pubKeyPath
     this.privKeyPath = privKeyPath
     this.disableAuth = disableAuth
+    this.rpcUser = rpcUser
+    this.rpcPass = rpcPass
+    this.auth = basicAuth.verify.bind(this)
```
treygriffith: why are you binding to the current context? Is there any reason that you need to?

broker-daemon/utils/basic-auth.js
```diff
@@ -0,0 +1,44 @@
...
+ * @param {Object} metadata
+ * @param {Object} logger
+ * @param {Boolean} this.disableAuth
+ * @param {String} this.rpcUser
+ * @param {String} this.rpcPass
+ * @return {True} true - if the user is authenticated
+ * @return {Error} if the user is not authenticated
+ */
+function verify ({ metadata, logger }) {
+  if (this.disableAuth === true) return
```
treygriffith: If we can avoid using context I think we should

broker-daemon/utils/basic-auth.js
```diff
@@ -0,0 +1,44 @@
...
+  logger.debug('Received auth token', { scheme, base64Token })
+
+  const rawCredentials = Buffer.from(base64Token, 'base64').toString()
+  const [username, password] = rawCredentials.split(BASIC_AUTH_DELIMITER)
+
+  logger.debug('Checking if credentials are valid between cli and broker')
+
+  // rpcUser and rpcPass are bound to context where #verify is used due to the way
+  // grpc-method handles the authorization middleware
+  if (username !== this.rpcUser || password !== this.rpcPass) {
```
treygriffith: see previous comment - if we can avoid using implicit context I think we should.

Instead, we could do something like: `createAuth(user, pass)` that returns a function.

docker-compose.yml
dannypaz: DISABLE_AUTH is now for broker auth where as DISABLE_RELAYER_AUTH is used for the relayer value

broker-daemon/utils/basic-auth.js
dannypaz: ill have to think about this. The issue is that verify is called from within grpc-methods which has no information/data for these parameters (which is why we use bind above)

broker-daemon/broker-rpc/broker-rpc-server.js
dannypaz: @treygriffith we bind to the current context because `verify` is called from within grpc-methods but is unable to access (or pass in arguments) for user/pass

docker-compose.yml
treygriffith: Cool - why is broker auth disabled by default? Is it a "until we get docs in place" thing?

broker-daemon/utils/basic-auth.js
treygriffith: you can instantiate within the RPC Server constructor and still pass a dumb method to grpc-methods.

```
   const auth = createAuth(user, pass)
   this.adminService = new AdminService(..., auth)
```

broker-cli/utils/basic-auth.js
dannypaz: I agree. its a bad name

docker-compose.yml
dannypaz: @treygriffith until its an official release

broker-daemon/broker-rpc/broker-rpc-server.js
```diff
@@ -79,21 +52,22 @@ class BrokerRPCServer {
     this.pubKeyPath = pubKeyPath
     this.privKeyPath = privKeyPath
     this.disableAuth = disableAuth
+    this.auth = verifyBasicAuth(rpcUser, rpcPass, disableAuth)
```
treygriffith: [nit] I wouldn't call this "verify" since that seems to indicate that it is checking auth in this call. Maybe something like `basicAuth` or `createBasicAuth`?

broker-daemon/broker-rpc/broker-rpc-server.js
```diff
@@ -104,10 +78,41 @@ class BrokerRPCServer {
...
+    const key = readFileSync(this.privKeyPath)
+    const cert = readFileSync(this.pubKeyPath)
+
+    this.logger.debug(`Securing gRPC connections with SSL: key: ${this.privKeyPath}, cert: ${this.pubKeyPath}`)
+
+    return grpc.ServerCredentials.createSsl(
+      null, // no root cert needed for server credentials
+      [{
+        private_key: key,
+        // TODO: build a real cert chain instead of just using the root cert
```
treygriffith: do we ever expect to build a cert chain for the broker?

dannypaz: This was copy pasta. I dont think we expect the broker to need anything more than pub/private for now

broker-daemon/broker-rpc/broker-rpc-server.js
dannypaz: I would prefer `createBasicAuth`

broker-daemon/utils/create-basic-auth.js
```diff
@@ -0,0 +1,44 @@
...
+ * called from inside the grpc-methods authorization hook.
+ *
+ * @see {@link https://github.com/sparkswap/grpc-methods}
+ * @param {String} rpcUser
+ * @param {String} rpcPass
+ * @param {Boolean} [disableAuth=false]
+ * @return {Function}
+ */
+function createBasicAuth (rpcUser, rpcPass, disableAuth = false) {
+  return async ({ metadata, logger }) => {
```
treygriffith: why does this return an `async` function? It looks synchronous to me. Is this reserving optionality for future changes that may have an async credentials check (e.g. that does IO)?

dannypaz: @treygriffith this is `awaited` in grpc-methods

dannypaz:
@treygriffith auth is called inside of grpc-methods.

PR #239, opened 2018-08-16
update createReadStream to take in updated range options

martineehrlich:
## Description
When a user submits a limit order to the broker, it should fill any orders that are at the limit price or better. Currently, if there is an order on the book at the exact same price, the broker won't fill it.

The issue was that we were only passing the quantumPrice to createReadStream in streamOrdersAtPriceOrBetter:

```
  streamOrdersAtPriceOrBetter (quantumPrice) {
    const opts = {}

    if (quantumPrice) {
      opts.lte = this.keyForPrice(quantumPrice)
    }
    return this.createReadStream(opts)
  }
```

createReadStream in sublevel index just passed these options through instead of getting updated in the range method, which means that any orders with the same quantum price were always seen as greater than the lte value:

`opts.lte = '00000000000000000000.0010000000000000000'`

All orders with that quantumPrice look like: `00000000000000000000.0010000000000000000:asdjhfakjsdgag`

So we were making the wrong comparison.

The fix is to call this.range on the options before passing the options into createReadStream so it actually takes our new range values into account:

After calling this.range on the options
`opts.lte = '00000000000000000000.0010000000000000000:\uffff'`

So that any order with the same quantum price will be less than or equal to opts.lte.

I used object.assign so that we don't directly mutate the options. 

## Todos
- [x] Tests
- [x] Documentation
- [x] Link to Trello

### DIFF
--- a/broker-daemon/utils/sublevel-index.js
+++ b/broker-daemon/utils/sublevel-index.js
@@ -94,7 +94,9 @@ class Index {
    * @return {Readable}    Readable stream
    */
   createReadStream (opts) {
-    const stream = this._index.createReadStream(opts)
+    const optionsToUpdate = Object.assign({}, opts)
+    const updatedOptions = this.range(optionsToUpdate)
+    const stream = this._index.createReadStream(updatedOptions)
     // through2 - the lib used below - overrides function context to provide access to `this.push` to push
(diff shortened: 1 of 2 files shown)


martineehrlich:
[limit orders don't pick up orders at the exact same price](https://trello.com/c/zkpkmYUg/617-limit-orders-dont-pick-up-orders-at-the-exact-same-price)

PR #243, opened 2018-08-30
FEATURE - complete block orders

dannypaz:
## Description
This PR adds functionality to the broker that allows us to take a block order into a `COMPLETED` status once all orders have been fulfilled.

Changes:
1. added events for completed/rejected
2. Created a new state machine plugin called `StateMachineEvents`
3. used EventEmitters for events in a order to communicate with a block order through the use of `StateMachineEvents`
4. updated block order to only set its complete status if all orders/fills are fulfilled
5. tests

## Related PRs

## Todos
- [ ] Tests
- [ ] Documentation
- [x] Link to Trello


dannypaz:
[Change order status to COMPLETED when it's completed [CCXT]](https://trello.com/c/fBfocb4m/635-change-order-status-to-completed-when-its-completed-ccxt)

dannypaz:
[inform clients of orders that are cancelled bc the relayer went down](https://trello.com/c/NjzCIgwD/620-inform-clients-of-orders-that-are-cancelled-bc-the-relayer-went-down)

broker-cli/certs/.gitignore
```diff
@@ -1,4 +1,5 @@
-# Ignore everything in this directory except this file
+# Ignore everything in this directory except this file and a README
```
treygriffith: I think this is the wrong pr to make this change in.

broker-cli/certs/.gitignore
```diff
@@ -1,4 +1,5 @@
...
 # We never want to inadvertantly check-in files to the certs/ directory
+*
```
treygriffith: Don't make this change in this PR.

broker-cli/test/test-helper.js
```diff
@@ -9,7 +9,10 @@
 // unhandled promise rejections are not handled by default in mocha currently,
 // but are deprecated in node.js, so should indicate test failure
 // @see {@link https://github.com/mochajs/mocha/issues/2640|mochajs/mocha#2640}
-process.on('unhandledRejection', () => { throw new Error('Unhandled rejection during testing') })
+process.on('unhandledRejection', (err) => {
+  console.error(err)
```
treygriffith: I don't think this error is the one that you want. Have you tested this?

broker-daemon/block-order-worker/index.js
```diff
@@ -1,17 +1,25 @@
...
 const { promisify } = require('util')
 const safeid = require('generate-safe-id')
+
 const { BlockOrder, Order, Fill } = require('../models')
 const { OrderStateMachine, FillStateMachine } = require('../state-machines')
+const {
+  Big,
+  getRecords,
+  SublevelIndex,
+  events: { BlockOrderWorkerEvents }
```
treygriffith: I don't think this should be in utils since it's specific to this module

broker-daemon/block-order-worker/index.js
```diff
@@ -91,7 +126,26 @@ class BlockOrderWorker extends EventEmitter {
 
     this.logger.info(`Created and stored block order`, { blockOrderId: blockOrder.id })
 
-    this.emit('BlockOrder:create', blockOrder)
+    // Register listener events by the current blockOrderId
+    this.once(BlockOrderWorkerEvents.COMPLETE + id, async (blockOrderId) => this.completeBlockOrder(blockOrderId))
```
treygriffith: I don't like the concatenation here, I'd rather this be a function I think.

broker-daemon/models/block-order.js
```diff
@@ -160,7 +160,24 @@ class BlockOrder {
...
+   * @return {BlockOrder} Modified block order instance
+   */
+  complete () {
+    const activeOrders = this.openOrders.filter(o => o.status !== BlockOrder.STATUSES.COMPLETED)
+    const activeFills = this.fills.filter(f => f.status !== BlockOrder.STATUSES.COMPLETED)
+
+    // If we still have active orders or fills, then the block order is not completed
+    // so we can simply return the object
+    if (activeOrders || activeFills) {
+      return this
```
treygriffith: It seems odd to not error out if we aren't updating the status. Why do that?

broker-daemon/models/block-order.js
```diff
@@ -160,7 +160,24 @@ class BlockOrder {
...
     this.status = BlockOrder.STATUSES.FAILED
+    return this
+  }
+
+  /**
+   * Move the block order to a completed status
+   * @return {BlockOrder} Modified block order instance
+   */
+  complete () {
+    const activeOrders = this.openOrders.filter(o => o.status !== BlockOrder.STATUSES.COMPLETED)
```
treygriffith: Can't a block order be completed even if an underlying order was cancelled?

broker-daemon/models/block-order.js
```diff
@@ -160,7 +160,24 @@ class BlockOrder {
...
+    const activeFills = this.fills.filter(f => f.status !== BlockOrder.STATUSES.COMPLETED)
```
treygriffith: See comment above: can't a block order be completed even if a fill was rejected (shouldn't we even expect that case)?

broker-daemon/state-machines/fill-state-machine.js
```diff
@@ -111,10 +114,11 @@ const FillStateMachine = StateMachine.factory({
    * @param  {RelayerClient}       options.relayer
    * @param  {Map<String, Engine>} options.engines     Collection of all avialable engines
    * @param  {Function}            options.onRejection A function to handle rejections of the fill
+   * @param  {Function}            options.onCompletion A function to handle the completion of the fill
    * @return {Object}                                  Data to attach to the state machine
    */
-  data: function ({ store, logger, relayer, engines, onRejection = function () {} }) {
-    return { store, logger, relayer, engines, onRejection, fill: {} }
+  data: function ({ store, logger, relayer, engines, worker }) {
```
treygriffith: I'm not a fan of exposing the block order worker to the state machine. I think the worker should know about the state machine, but not the inverse of that. I don't think there is a need to make that relationship two-way.

broker-daemon/block-order-worker/index.js
dannypaz: it is shared between modules

broker-daemon/models/block-order.js
dannypaz: This is a future change. If the block order still has active orders/fills we dont want to update its status to completed

broker-daemon/models/block-order.js
dannypaz: Yup. im missing this case

broker-daemon/state-machines/fill-state-machine.js
dannypaz: @treygriffith its not necessarily exposing the worker as much as exposing the events API. The current use of `onRejection` was ok, but its real goal is message passing. The only reason we use `worker` here is for triggering an event that the block order is listening for, that the block order can only set.

dannypaz: @treygriffith This was the tradeoff that I made vs using a callback. I tried to avoid this situation because we both had agreed earlier that we wanted the worker/state machine separation. This was one of the cleanest implementation choices

broker-cli/test/test-helper.js
dannypaz: @treygriffith I had legit test failures that were being swallowed by this call. We should log the error JIC

dannypaz:
After talking w/ @treygriffith we have decided to have the state machines be event emitters, where the block order worker will add events directly on the state machine in order to pass messages between the two.

broker-daemon/block-order-worker/index.js
dannypaz: @treygriffith we are removing this in the next iteration of the PR

broker-daemon/block-order-worker/index.spec.js
```diff
@@ -1090,25 +1088,14 @@ describe('BlockOrderWorker', () => {
...
+    it('registers an event on an order for order completion', async () => {
       await worker._placeOrder(blockOrder, '100')
-
-      expect(OrderStateMachine.create).to.have.been.calledWith(sinon.match({ onRejection: sinon.match.func }))
+      expect(onceStub).to.have.been.calledWith('complete', sinon.match.func)
     })
 
-    it('fails the block order if the order is rejected', async () => {
-      worker.failBlockOrder = sinon.stub()
+    it('registers an event on an order for order rejection', async () => {
```
treygriffith: don't you need to test the function that is registered as well?

broker-daemon/models/block-order.js
```diff
@@ -159,8 +159,26 @@ class BlockOrder {
...
+   * @return {BlockOrder} Modified block order instance
+   */
+  complete () {
+    const activeOrders = this.openOrders.filter(o => o.status === BlockOrder.STATUSES.ACTIVE)
+    const activeFills = this.fills.filter(f => f.status === BlockOrder.STATUSES.ACTIVE)
+
+    // If we still have active orders or fills, then the block order is not completed
+    // so we can simply return the object
+    if (activeOrders.length > 0 || activeFills.length > 0) {
+      return this
```
treygriffith: shouldn't this be an error state of some kind?

broker-daemon/relayer/relayer-client.spec.js
```diff
@@ -36,6 +37,11 @@ describe('RelayerClient', () => {
   let ENABLE_SSL = true
 
   beforeEach(() => {
+    logger = {
```
treygriffith: can this be in another pr? seems unrelated to this change.

broker-daemon/state-machines/plugins/state-machine-events.js
```diff
@@ -0,0 +1,69 @@
...
+
+/**
+ * Events that should trigger the cleanup of all event handlers created
+ * by the StateMachineEvents plugin
+ *
+ * @constant
+ * @type {Array<String>}
+ * @default
+ */
+const EXIT_EVENTS = Object.freeze(['cancel', 'complete', 'reject'])
```
treygriffith: I think these plugins should be generic, something we would be comfortable using for any state machine.

This should be pulled into configuration for the plugin.

broker-daemon/state-machines/plugins/state-machine-events.js
```diff
@@ -0,0 +1,69 @@
...
+
+    return {
+      /**
+       * @param {Object} lifecycle
+       * @returns {void}
+       */
+      onAfterTransition: function (lifecycle) {
+        // We check to see if an order exists, if it does then we are dealing with an
+        // OSM, or else we use the FSM
+        const blockOrderId = this.order ? this.order.blockOrderId : this.fill.blockOrderId
```
treygriffith: the state machine should not have knowledge of the block order, and this plugin should not have knowledge of how the order and fill state machines are structured.

We shouldn't be emitting the block order id at all - the block order is itself listening to this event emitter, so the ID is superfluous. Emitting the lifecycle event is plenty specific enough.

broker-daemon/state-machines/plugins/state-machine-events.js
```diff
@@ -0,0 +1,69 @@
...
+  }
+
+  /**
+   * Alias methods on a StateMachine for event handling
+   */
+  get methods () {
+    const { eventHandler } = this
+
+    return {
+      once: (type, cb) => eventHandler.once(type, cb)
```
treygriffith: why expose only `once`? why not provide the full API? e.g. `on`, `once`, `removeListener`, etc

broker-daemon/block-order-worker/index.js
```diff
@@ -126,26 +114,22 @@ class BlockOrderWorker extends EventEmitter {
...
-      this.removeAllListeners(BlockOrderWorkerEvents.CREATED + id)
-    })
+    // Move BlockOrder to a rejected state if any process in working a block order, fails
+    this.once(rejectedEventId, async (blockOrderId, err) => this.failBlockOrder(blockOrderId, err))
 
-    this.on(BlockOrderWorkerEvents.CREATED + id, async (blockOrder) => {
-      await this.workBlockOrder(blockOrder).catch(e => this.emit(BlockOrderWorkerEvents.REJECTED, blockOrder.id, e))
+    // Start working the block order in another process to prevent blocking the creation
+    // of 'other' block orders
+    this.once(createEventId, async (blockOrder) => {
```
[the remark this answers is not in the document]
dannypaz: @treygriffith I don't think there is any reason that would be an issue.

broker-daemon/block-order-worker/index.js
```diff
@@ -1,17 +1,45 @@
...
+ * Event names used for handling emitted events during the 'working' of a single
+ * block order
+ *
+ * @constant
+ * @type {Object}
+ * @default
+ */
+const WORKER_EVENTS = Object.freeze({
+  CREATE: 'CREATE',
+  REJECTED: 'REJECTED'
```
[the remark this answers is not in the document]
dannypaz: @treygriffith this was a part of the original implementation. We are going to remove the `CREATE` event

broker-daemon/block-order-worker/index.js
```diff
@@ -60,7 +81,9 @@ class BlockOrderWorker extends EventEmitter {
   }
 
   /**
-   * Create a new block order
+   * Creates a new block order and registers events for all orders under a block order
+   *
+   * @param {Object} options
```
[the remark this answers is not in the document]
dannypaz: @treygriffith you'll need to find a source. This hasn't been the case in my experience. If we omit `@param {Object} options` then you'll have parent properties that say `options.whatever` w/ out proper nesting in the template.

broker-daemon/block-order-worker/index.js
```diff
@@ -91,7 +114,22 @@ class BlockOrderWorker extends EventEmitter {
...
+    // Register events on a block order worker to handle processing of concurrent orders
+    const createEventId = generateIdForEvent(WORKER_EVENTS.CREATE, id)
+    const rejectedEventId = generateIdForEvent(WORKER_EVENTS.REJECTED, id)
+
+    // Move BlockOrder to a rejected state if any process in working a block order, fails
+    this.once(rejectedEventId, async (blockOrderId, err) => this.failBlockOrder(blockOrderId, err))
+
+    // Start working the block order in another process to prevent blocking the creation
+    // of 'other' block orders
+    this.once(createEventId, async (blockOrder) => {
```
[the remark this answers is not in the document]
dannypaz: Dupe comment. see reply above

broker-daemon/models/block-order.js
dannypaz: I made a comment above. We shouldn't complete an BlockOrder if there are existing active fills or orders. This is assuming that a BlockOrder -> has many orders/fills.

broker-daemon/block-order-worker/index.js
```diff
@@ -345,15 +406,20 @@ class BlockOrderWorker extends EventEmitter {
...
-        }
+        store
       },
       blockOrder.id,
       { side, baseSymbol, counterSymbol, baseAmount, counterAmount }
     )
 
+    // Register listener events for the current order
+    const onceCompleteEvent = async (blockOrderId) => this.completeBlockOrder(blockOrderId)
+    const onceRejectedEvent = async (blockOrderId) => this.failBlockOrder(blockOrderId, order.error)
```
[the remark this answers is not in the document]
dannypaz: Doesnt need to be. works all the same

broker-daemon/block-order-worker/index.js
```diff
@@ -345,15 +406,20 @@ class BlockOrderWorker extends EventEmitter {
...
-          this.failBlockOrder(blockOrder.id, err)
-        }
+        store
       },
       blockOrder.id,
       { side, baseSymbol, counterSymbol, baseAmount, counterAmount }
     )
 
+    // Register listener events for the current order
+    const onceCompleteEvent = async (blockOrderId) => this.completeBlockOrder(blockOrderId)
```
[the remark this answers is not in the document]
dannypaz: Doesnt need to be. works all the same

broker-daemon/block-order-worker/index.js
```diff
@@ -222,12 +256,10 @@ class BlockOrderWorker extends EventEmitter {
    * @return {void}
    */
   async failBlockOrder (blockOrderId, err) {
-    // TODO: expose the stack trace of err because it is hard to troubleshoot without it
-    this.logger.error('Error encountered while working block', { id: blockOrderId, error: err.message })
+    this.logger.error('Error encountered while working block', { id: blockOrderId, error: err.stack })
```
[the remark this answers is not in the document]
dannypaz: @treygriffith there was a comment that said `// TODO: include stack information because this error is hard to debug`, so I added the full stack

broker-daemon/block-order-worker/index.js
```diff
@@ -91,7 +114,22 @@ class BlockOrderWorker extends EventEmitter {
...
+    // Move BlockOrder to a rejected state if any process in working a block order, fails
+    this.once(rejectedEventId, async (blockOrderId, err) => this.failBlockOrder(blockOrderId, err))
+
+    // Start working the block order in another process to prevent blocking the creation
+    // of 'other' block orders
+    this.once(createEventId, async (blockOrder) => {
+      await this.workBlockOrder(blockOrder).catch(e => this.emit(rejectedEventId, blockOrder.id, e))
+    })
+
+    // TODO: Use an ID instead of the entire BlockOrder so that we can have multiple
```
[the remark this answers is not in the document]
dannypaz: I think its a good practice to pass IDs around instead of objects because the object can be mutated before the event occurs (which could lead to issues). ill update the comment

broker-daemon/block-order-worker/index.spec.js
```diff
@@ -202,45 +203,12 @@ describe('BlockOrderWorker', () => {
       expect(Order.fromStorage).to.have.been.calledWith(fakeKey, fakeValue)
       expect(filtered).to.be.equal(false)
     })
-
-    it('works a block order when one is created', async () => {
```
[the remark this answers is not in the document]
dannypaz: Test was removed from this section of the spec because we no longer use events in the initialization of a BlockOrderWorker. we test for this below.

broker-daemon/block-order-worker/index.spec.js
```diff
@@ -202,45 +203,12 @@ describe('BlockOrderWorker', () => {
...
-      worker = new BlockOrderWorker({ orderbooks, store, logger, relayer, engines })
-      worker.workBlockOrder = sinon.stub().resolves()
-
-      await delay(15)
-
-      expect(worker.workBlockOrder).to.have.been.calledOnce()
-      expect(worker.workBlockOrder).to.have.been.calledWith(fakeBlockOrder)
-    })
-
-    it('fails a block order if working it fails', async () => {
```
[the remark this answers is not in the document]
dannypaz: Same as the above test. This test was removed from this section of the spec because we no longer use events in the initialization of a BlockOrderWorker. we test for this below.

broker-daemon/block-order-worker/index.spec.js
dannypaz: Yes. I will add the test. thanks for the catch

broker-daemon/relayer/relayer-client.spec.js
dannypaz: @treygriffith im not entirely sure where this change came from, but it shouldnt exist in the diff anymore (after updating with master)

broker-daemon/state-machines/plugins/state-machine-events.js
dannypaz: @treygriffith I've omitted those other events because they are not used

broker-daemon/state-machines/plugins/state-machine-events.js
dannypaz: @treygriffith I totally agree. I've removed the need for an ID during event handling. When these events are registered with a state machine, the caller should be providing the information needed. 

This code no longer exists

broker-daemon/block-order-worker/index.js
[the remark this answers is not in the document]
dannypaz: FWIW `err.stack` includes the message

broker-daemon/state-machines/plugins/state-machine-events.js
treygriffith: Ok, let's punt for now

broker-daemon/block-order-worker/index.js
```diff
@@ -316,6 +312,37 @@ class BlockOrderWorker extends EventEmitter {
...
+  /**
+   * Move a block order to a completed state if all orders have been completed
+   * @param {String} blockOrderId
+   */
+  async completeBlockOrder (blockOrderId) {
+    this.logger.info('Attempting to put block order in a completed state', { id: blockOrderId })
+
+    // TODO: move status to its own sublevel so it can be updated atomically
+    try {
+      var value = await promisify(this.store.get)(blockOrderId)
```
treygriffith: since we do this twice, we should consider moving it to its own method

broker-daemon/block-order-worker/index.js
```diff
@@ -316,6 +312,37 @@ class BlockOrderWorker extends EventEmitter {
...
+      }
+
+      throw e
+    }
+
+    const blockOrder = BlockOrder.fromStorage(blockOrderId, value)
+
+    // An order is only completed if all orders underneath the blockorder are out of
+    // an `ACTIVE` state.
+    blockOrder.complete()
```
treygriffith: per our discussion, any validity checks should happen here (not in the model).

Also, we should check the actual amount of the block order and not just the underlying orders/fills.

broker-daemon/state-machines/plugins/state-machine-events.js
```diff
@@ -0,0 +1,56 @@
...
+       * @param {Object} lifecycle
+       * @returns {void}
+       */
+      onAfterTransition: function (lifecycle) {
+        plugin.eventHandler.emit(lifecycle.transition)
+
+        // Remove all events from the event emitter if we have either cancelled,
+        // completed, or rejected an order
+        if (plugin.exitEvents.includes(lifecycle.transition)) {
+          plugin.eventHandler.removeAllListeners()
```
treygriffith: this seems strange. Why do we need this? And why do it this way?

dannypaz: @treygriffith just a cleanup of events after we mark a StateMachine as "finished". Would hopefully free up some memory or there would be a leak

broker-daemon/block-order-worker/index.js
```diff
@@ -316,6 +281,57 @@ class BlockOrderWorker extends EventEmitter {
...
+
+    const blockOrder = await this.getBlockOrder(blockOrderId)
+
+    // check the fillamount on each collection of state machines from the block order
+    // and make sure that either is equal to how much we are trying to fill.
+    let totalFilled = Big(0)
+
+    if (blockOrder.fills.length) {
+      totalFilled = blockOrder.fills.reduce((acc, fsm) => acc.plus(fsm.fill.fillAmount), Big(0))
+    } else if (blockOrder.openOrders.length) {
```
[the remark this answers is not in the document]
dannypaz: @treygriffith what is the situation where a single blockorder would have both of these?

broker-daemon/block-order-worker/index.js
```diff
@@ -316,6 +281,57 @@ class BlockOrderWorker extends EventEmitter {
...
+
+    if (!activeOrders && !activeFills && !stillBeingFilled) {
+      // An order is only completed if all orders underneath the blockorder are out of
+      // an `ACTIVE` state and it is entirely filled
+      blockOrder.complete()
+      await promisify(this.store.put)(blockOrder.key, blockOrder.value)
+      this.logger.info('Moved block order to completed state', { blockOrderId })
+    } else if (stillBeingFilled && !activeFills && !activeOrders) {
+      this.logger.error('Block order is still being filled, but has no active orders/fills', { blockOrderId })
+      throw new Error('Block order in a weird state. Order is filled, but no active orders/fills still exist')
```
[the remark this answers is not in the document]
dannypaz: @treygriffith thats why we throw the error. If its in a weird state, we want to know and need to fix the bug. I dont think we should avoid this state because of a race-condition. We can simply re-run this method and the order will effectively complete, but it is still nice to know if this situation occurs.

Either way, if I should remove the check, that is fine too.

broker-daemon/block-order-worker/index.js
```diff
@@ -316,6 +281,57 @@ class BlockOrderWorker extends EventEmitter {
...
+    } else if (!stillBeingFilled && (activeFills || activeOrders)) {
+      this.logger.error('Block order is not being filled and has active orders/fills', { activeFills, activeOrders, blockOrderId })
+      throw new Error('Block order in a weird state. Order is not being filled, but active orders/fills still exist')
```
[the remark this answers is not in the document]
dannypaz: @treygriffith I think this case is less of an issue than the one above. If the order is not being filled (meaning the fillAmounts match), then we would have effectively completed all orders (the property `active` would HAVE to be false). If that property is never switched, then we have a bug that we need to be aware of. I dont see an edgecase for this piece of code.

broker-daemon/state-machines/order-state-machine.js
```diff
@@ -362,15 +383,7 @@ const OrderStateMachine = StateMachine.factory({
...
-    },
-
-    /**
-     * Handle rejected state by calling a passed in handler
-     * @param  {Object} lifecycle Lifecycle object passed by javascript-state-machine
-     * @return {void}
-     */
-    onAfterReject: async function (lifecycle) {
-      this.onRejection(this.error)
+      this.order.error = error
```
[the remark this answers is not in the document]
dannypaz: @treygriffith yeah, we lost it with the removal of the method call

broker-daemon/block-order-worker/index.js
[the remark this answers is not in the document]
dannypaz: @treygriffith I've had a second thought. If this piece of code makes the `complete` method too complicated (or weird) then we should remove it. If this issue occurs, it is probably better to realize and error at a different point in the lifecycle, than at the very end. What do you think?

broker-daemon/models/order.js
```diff
@@ -246,6 +246,16 @@ class Order {
...
   }
 
+  /**
+   * @return {Boolean}
+   */
+  get active () {
+    // We currently define an "ACTIVE" order as an Order w/out a fillAmount, since this
+    // property is only set once an order has been filled.
+    // TODO: Figure out a better definition of an "ACTIVE" order (setting a status?)
+    return (!this.swapPreimage)
```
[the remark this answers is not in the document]
dannypaz: ty good catch @treygriffith

broker-daemon/block-order-worker/index.js
```diff
@@ -397,21 +419,28 @@ class BlockOrderWorker extends EventEmitter {
...
       )
+
+      // The events below tie into the StateMachine's lifecycle hooks for a OrderStateMachine
+      fsm.once('execute', () => {
+        this.completeBlockOrder(blockOrder.id).catch(e => {
+          this.logger.error(`BlockOrder failed to be completed, status is left active: ${blockOrder.id}`, { error: e.stack })
+        })
+      })
+
+      fsm.once('reject', () => this.failBlockOrder(blockOrder.id, fsm.fill.error))
```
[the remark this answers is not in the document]
dannypaz: Totally agree. At the very least it'd be nice to log that we failed to fail

broker-daemon/block-order-worker/index.js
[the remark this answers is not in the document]
treygriffith: limit order where there are existing orders on the books at that price

broker-daemon/block-order-worker/index.js
```diff
@@ -316,6 +281,43 @@ class BlockOrderWorker extends EventEmitter {
...
+      return acc.plus(osm.order.fillAmount || 0)
+    }, totalFilled)
+
+    this.logger.debug('Current total filled amount: ', { totalFilled, blockOrderAmount: blockOrder.baseAmount })
+
+    const stillBeingFilled = totalFilled.lt(blockOrder.baseAmount)
+    const activeOrders = blockOrder.openOrders.some(osm => osm.order.active)
+    const activeFills = blockOrder.fills.some(fsm => fsm.fill.active)
+
+    if (!activeOrders && !activeFills && !stillBeingFilled) {
```
treygriffith: if it is not still being filled and it has active orders, what behavior do we want?

broker-daemon/block-order-worker/index.js
```diff
@@ -316,6 +281,43 @@ class BlockOrderWorker extends EventEmitter {
...
+      // An order is only completed if all orders underneath the blockorder are out of
+      // an `ACTIVE` state and it is entirely filled
+      blockOrder.complete()
+      await promisify(this.store.put)(blockOrder.key, blockOrder.value)
+      this.logger.info('Moved block order to completed state', { blockOrderId })
+    } else {
+      this.logger.debug('Block order is not ready to be completed', { blockOrderId })
```
treygriffith: should this be an error?

broker-daemon/block-order-worker/index.js
```diff
@@ -340,21 +342,38 @@ class BlockOrderWorker extends EventEmitter {
...
 
-    this.logger.info('Created order for BlockOrder', { blockOrderId: blockOrder.id, orderId: order.orderId })
+    // We are hooking into the complete lifecycle event of an order state machine to trigger
+    // the completion of a blockorder
+    osm.once('complete', async () => {
+      this.completeBlockOrder(blockOrder.id)
+        .catch(e => {
+          this.logger.error(`BlockOrder failed to be completed from order`, { id: blockOrder.id, error: e.stack })
+        })
+        .then(() => osm.removeAllListeners())
```
treygriffith: if we are using a `once`, do we need to clean up listeners here?

broker-daemon/models/fill.js
```diff
@@ -243,6 +243,16 @@ class Fill {
...
   }
 
+  /**
+   * @return {Boolean}
+   */
+  get active () {
+    // We define an "ACTIVE" fill as a Fill that has not been assigned a maker address
+    // meaning that the fill has no been set to execute.
+    // TODO: Figure out a better definition of an "ACTIVE" fill (setting a status?)
+    return (!this.makerAddress)
```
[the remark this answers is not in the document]
treygriffith: Adding on to this, it may make sense to make sure the fill hasn't been rejected

broker-daemon/models/order.js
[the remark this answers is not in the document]
treygriffith: This still needs to be fixed @dannypaz

broker-daemon/block-order-worker/index.spec.js
treygriffith: @dannypaz where is the test for the registered functions?

broker-daemon/block-order-worker/index.js
dannypaz: @treygriffith we intentionally removed checks for that functionality

broker-daemon/block-order-worker/index.js
dannypaz: Not necessarily. There are cases where this would be an error, however we decided to remove them. As an example: This would be hit if the block order is still being filled, which is a valid case and wouldn't be an error

broker-daemon/block-order-worker/index.js
dannypaz: We still need to clean up, because we would have a `reject` listener still open on the `osm`

broker-daemon/models/order.js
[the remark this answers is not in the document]
dannypaz: @treygriffith we decided to punt this. there is a ticket in the backlog for adding a `status` to an order.

broker-daemon/block-order-worker/index.js
```diff
@@ -340,21 +342,40 @@ class BlockOrderWorker extends EventEmitter {
...
+
+    // We are hooking into the reject lifecycle event of an order state machine to trigger
+    // the failure of a blockorder
+    osm.once('reject', async () => {
+      this.failBlockOrder(blockOrder.id, osm.order.error)
+        .catch(e => {
+          this.logger.error(`BlockOrder failed on setting a failed status from order`, { id: blockOrder.id, error: e.stack })
+        })
+        .then(() => {
+          return osm.removeAllListeners()
```
treygriffith: [nit] weird that you did multiple lines on this one but single line for `complete`

broker-daemon/models/block-order.js
```diff
@@ -263,6 +276,30 @@ class BlockOrder {
...
+   * Grab a block order from a given sublevel
+   *
+   * @param {Sublevel} store block order sublevel store
+   * @param {String} blockOrderId
+   * @return {BlockOrder} BlockOrder instance
+   * @throws {Error} store is null
+   * @throws {BlockOrderNotFoundError} block order could not be found
+   */
+  static async fromStore (store, blockOrderId) {
+    if (!store) throw new Error('[BlockOrder#fromStore] No leveldb store is defined')
```
treygriffith: [nit] I don't think there is a good reason to reference leveldb in the error message

broker-daemon/models/order.js
```diff
@@ -246,6 +246,16 @@ class Order {
     }
   }
 
+  /**
+   * @return {Boolean}
+   */
+  get active () {
```
treygriffith: why define this in this PR if we aren't using it?

broker-daemon/block-order-worker/index.spec.js
```diff
@@ -885,14 +794,140 @@ describe('BlockOrderWorker', () => {
...
+            { order: { fillAmount: 100 } }
+          ]
+        }
+        getBlockOrderStub.returns(activeBlockOrder)
+        await worker.checkBlockOrderCompletion(blockOrder.id)
+        expect(blockOrderCompleteStub).to.not.have.been.calledOnce()
+      })
+    })
+
+    context('block order with fills', () => {
```
treygriffith: consider adding a test with a block order with both orders and fills

broker-daemon/block-order-worker/index.spec.js
```diff
@@ -885,14 +794,140 @@ describe('BlockOrderWorker', () => {
...
+        worker.blockOrder = blockOrder
+        worker.getBlockOrder = getBlockOrderStub.returns(blockOrder)
+      })
+
+      it('completes a block order', async () => {
+        await worker.checkBlockOrderCompletion(blockOrder.id)
+        expect(blockOrderCompleteStub).to.have.been.calledOnce()
+      })
+
+      it('does not complete a block order if order is still active', async () => {
```
treygriffith: is this test mis-named? the underlying orders being active is no longer a condition

broker-daemon/models/order.js
dannypaz: Good catch. I forgot to remove these from the order/fill.

PR #244, opened 2018-09-06
FIX - Max Channel Balance per currency

dannypaz:
## Description
This PR addresses an issue with committing balances in LTC, where LTC will be capped at the BTC max channel balance (16777215) instead of the correct balance (1006632900).

The max channel balances are currently set in LND at https://github.com/lightningnetwork/lightning-rfc/blob/master/02-peer-protocol.md#rationale

BTC is 2^24
LTC is BTC * 60 ([btcToLtcConversionRate)](https://github.com/lightningnetwork/lnd/blob/f1256ba41b033e62de2980b89bc7e278f76a0880/chainregistry.go#L57)

## Related PRs

## Todos
- [x] Tests
- [x] Documentation
- [x] Link to Trello

### DIFF
--- a/broker-cli/commands/wallet.js
+++ b/broker-cli/commands/wallet.js
@@ -5,8 +5,9 @@
 
 const Table = require('cli-table')
 require('colors')
+
 const BrokerDaemonClient = require('../broker-daemon-client')
-const { ENUMS, validations, askQuestion, Big, handleError } = require('../utils')
+const { validations, askQuestion, Big, handleError } = require('../utils')
 const { currencies: currencyConfig } = require('../configuration')
 
 /**
@@ -125,8 +126,13 @@ async function newDepositAddress (args, opts, logger) {
 async function commit (args, opts, logger) {
   const { symbol, amount } = args
   const { rpcAddress = null, market } = opts
+  const currentCurrencyConfig = currencyConfig.find(({ symbol: configSymbol }) => configSymbol === symbol)
 
   try {
+    if (!currentCurrencyConfig) {
+      throw new Error(`Currency is not supported by the CLI: ${symbol}`)
+    }
+
     const client = new BrokerDaemonClient(rpcAddress)
 
     const { balances } = await client.walletService.getBalances({})
@@ -140,24 +146,25 @@ async function commit (args, opts, logger) {
     }
 
     // We try to take the lowest total here between 2 Big numbers due to a
-    // commit limit specified as MAX_CHANNEL_BALANCE
+    // commit limit specified as `maxChannelBalance` in the currency configuration
     let maxSupportedBalance = uncommittedBalance
 
-    // TODO: Change the MAX_CHANNEL_BALANCE to be set per currency
-    if (totalUncommittedBalance.gt(ENUMS.MAX_CHANNEL_BALANCE)) {
-      maxSupportedBalance = ENUMS.MAX_CHANNEL_BALANCE
+    const maxChannelBalance = currentCurrencyConfig.maxChannelBalance
+
+    if (totalUncommittedBalance.gt(maxChannelBalance)) {
+      maxSupportedBalance = maxChannelBalance
     }
 
     // We use this for normalization of the amount from satoshis to a decimal
     // value
-    const divideBy = currencyConfig.find(({ symbol: configSymbol }) => configSymbol === symbol).quantumsPerCommon
+    const divideBy = currentCurrencyConfig.quantumsPerCommon
 
     // Amount is specified in currency instead of satoshis
     if (amount) {
       maxSupportedBalance = (amount * divideBy)
     }
 
-    logger.info(`For your knowledge, the Maximum supported balance at this time is: ${Big(ENUMS.MAX_CHANNEL_BALANCE).div(divideBy)} ${symbol}`)
+    logger.info(`For your knowledge, the Maximum supported balance at this time is: ${Big(maxSupportedBalance).div(divideBy)} ${symbol}`)
     logger.info(`Your current uncommitted wallet balance is: ${Big(uncommittedBalance).div(divideBy)} ${symbol}`)
 
     const answer = await askQuestion(`Are you OK committing ${Big(maxSupportedBalance).div(divideBy)} ${symbol} to sparkswap? (Y/N)`)
--- a/broker-cli/commands/wallet.js
+++ b/broker-cli/commands/wallet.js
@@ -147,9 +147,9 @@ async function commit (args, opts, logger) {
 
     // We try to take the lowest total here between 2 Big numbers due to a
     // commit limit specified as `maxChannelBalance` in the currency configuration
-    let maxSupportedBalance = uncommittedBalance
+    let maxSupportedBalance = totalUncommittedBalance
 
-    const maxChannelBalance = currentCurrencyConfig.maxChannelBalance
+    const maxChannelBalance = Big(currentCurrencyConfig.maxChannelBalance)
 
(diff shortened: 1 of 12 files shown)


dannypaz:
[Per currency for max channel balance on `commit-balance`](https://trello.com/c/VU9xQVnt/560-per-currency-for-max-channel-balance-on-commit-balance)

broker-daemon/broker-rpc/wallet-service/commit.js
```diff
@@ -1,25 +1,14 @@
...
- * Maximum channel balance (no inclusive) is 2^32 or 16777216
- * More info: https://github.com/lightningnetwork/lnd/releases/tag/v0.3-alpha
- *
- * @todo make this engine agnostic (non-LND)
  * @constant
  * @type {Long}
  * @default
  */
-const MAX_CHANNEL_BALANCE = 16777215
+const MINIMUM_FUNDING_AMOUNT = 400000
```
martineehrlich: What determines the minimum funding amount? Is it the same source as the max channel balance? If so why isn't this value also added to the configuration?

dannypaz: @martineehrlich minimum funding amount is a value that we set for channels. This change was not made in the PR (it shows up as green because I removed `MAX_CHANNEL_BALANCE` constant

broker-cli/commands/wallet.js
```diff
@@ -125,8 +126,13 @@ async function newDepositAddress (args, opts, logger) {
 async function commit (args, opts, logger) {
   const { symbol, amount } = args
   const { rpcAddress = null, market } = opts
+  const currentCurrencyConfig = currencyConfig.find(({ symbol: configSymbol }) => configSymbol === symbol)
```
treygriffith: I think we've had this conversation before, but why not just have the config inside of the daemon instead of the CLI? What's the drawback?

broker-cli/commands/wallet.js
```diff
@@ -140,31 +146,32 @@ async function commit (args, opts, logger) {
...
 
     // We use this for normalization of the amount from satoshis to a decimal
     // value
-    const divideBy = currencyConfig.find(({ symbol: configSymbol }) => configSymbol === symbol).quantumsPerCommon
+    const divideBy = currentCurrencyConfig.quantumsPerCommon
 
     // Amount is specified in currency instead of satoshis
     if (amount) {
-      maxSupportedBalance = (amount * divideBy)
+      maxSupportedBalance = Big(amount).times(divideBy)
```
treygriffith: can you add a test for this?

broker-daemon/broker-rpc/wallet-service/commit.js
```diff
@@ -61,14 +63,16 @@ async function commit ({ params, relayer, logger, engines, orderbooks }, { Empty
...
+
   // TODO: Validate that the amount is above the minimum channel balance
   // TODO: Choose the correct engine depending on the market
   // TODO: Get correct fee amount from engine
   if (balance < MINIMUM_FUNDING_AMOUNT) {
     throw new PublicError(`Minimum balance of ${MINIMUM_FUNDING_AMOUNT} needed to commit to the relayer`)
-  } else if (balance > MAX_CHANNEL_BALANCE) {
-    logger.error(`Balance from the client exceeds maximum balance allowed (${MAX_CHANNEL_BALANCE}).`, { balance })
-    throw new PublicError(`Maxium balance of ${MAX_CHANNEL_BALANCE} exceeded for committing to the relayer. Please try again.`)
+  } else if (balance > maxChannelBalance) {
```
treygriffith: should use big here instead of `>`

broker-daemon/broker-rpc/wallet-service/commit.js
```diff
@@ -61,14 +63,16 @@ async function commit ({ params, relayer, logger, engines, orderbooks }, { Empty
 
   logger.info(`Attempting to create channel with ${address} on ${symbol} with ${balance}`)
 
+  const maxChannelBalance = Number(currentCurrencyConfig.maxChannelBalance)
```
treygriffith: convert to Big instead of a number

broker-daemon/broker-rpc/wallet-service/commit.js
```diff
@@ -61,14 +63,16 @@ async function commit ({ params, relayer, logger, engines, orderbooks }, { Empty
...
+
   // TODO: Validate that the amount is above the minimum channel balance
   // TODO: Choose the correct engine depending on the market
   // TODO: Get correct fee amount from engine
   if (balance < MINIMUM_FUNDING_AMOUNT) {
```
treygriffith: should probably change this to Big

broker-daemon/broker-rpc/wallet-service/commit.js
```diff
@@ -96,11 +100,29 @@ async function commit ({ params, relayer, logger, engines, orderbooks }, { Empty
...
-  await engine.createChannel(address, balance)
+  logger.debug('Creating outbound channel')
+
+  try {
+    await engine.createChannel(address, balance)
+  } catch (e) {
+    logger.error('Received error when creating outbound channel', { e })
+
+    if (e.details && e.details.includes(CHANNEL_FUNDING_ERROR)) {
+      throw new PublicError(`Failed to commit wallet: Your wallet balance cannot cover on-chain fees. Please commit a smaller balance`)
```
treygriffith: is this the only way this error gets triggered? Being user-friendly is nice, but if there is a case that triggers this error that isn't the one we give to the user it will be very confusing.

broker-daemon/broker-rpc/wallet-service/commit.js
```diff
@@ -96,11 +100,29 @@ async function commit ({ params, relayer, logger, engines, orderbooks }, { Empty
...
+    }
+
+    throw (e)
+  }
 
   const paymentChannelNetworkAddress = await inverseEngine.getPaymentChannelNetworkAddress()
 
-  await relayer.paymentChannelNetworkService.createChannel({address: paymentChannelNetworkAddress, balance: convertedBalance, symbol: inverseSymbol})
+  try {
+    logger.debug('Requesting inbound channel from relayer')
```
treygriffith: including more info in this log would be nice

broker-daemon/broker-rpc/wallet-service/commit.js
```diff
@@ -96,11 +100,29 @@ async function commit ({ params, relayer, logger, engines, orderbooks }, { Empty
     throw new PublicError(errorMessage)
   }
 
-  await engine.createChannel(address, balance)
+  logger.debug('Creating outbound channel')
```
treygriffith: including more info in this log would be nice

broker-cli/commands/wallet.js
dannypaz: @treygriffith I would rather have it inside of the daemon. I think we have 2 separate configs because we wanted to ship the CLI as a separate entity. Do you know if we have a future item to change this?

treygriffith: @dannypaz but I suppose that means either: having the daemon expose an endpoint with this config, or not having the user feedback (e.g. "are you ok committing the max (which is XYZ BTC)?"). I don't want the CLI to reach into the daemon code to get the config.

I couldn't find a ticket for this issue.

broker-daemon/broker-rpc/wallet-service/commit.js
dannypaz: I dont like this either, because that error (string matching) can potentially change, but for now _i think_ this is OK. We have work coming up that will deal with fees that will make this error obsolete (for the most part).

PR #247, opened 2018-09-14
FIX - Resolve Orderbooks in bad state

dannypaz:
## Description
This PR addresses inconsistencies for orderbooks on the broker. Specifically it addresses an issue where the broker _may_ miss an event because a single timestamp can potentially have multiple events. To alleviate this, we have added a `eventOrder` field to the market event to differentiate between these events.

Changes Include:
1. adding a `eventOrder` field to market
2. Changed key used to store events in leveldb
3. added `getLastRecord` method to orderbook

## Related PRs

## Todos
- [ ] Tests
- [x] Documentation
- [x] Link to Trello

### DIFF
--- a/broker-daemon/models/market-event.js
+++ b/broker-daemon/models/market-event.js
@@ -1,5 +1,5 @@
 class MarketEvent {
-  constructor ({ eventId, orderId, timestamp, eventType, version, ...payload }) {
+  constructor ({ eventId, orderId, timestamp, eventType, eventOrder, ...payload }) {
     if (!Object.keys(this.constructor.TYPES).includes(eventType)) {
       throw new Error(`MarketEvents do not support a "${eventType}" type.`)
     }
@@ -8,13 +8,13 @@ class MarketEvent {
     this.orderId = orderId
     this.timestamp = timestamp
     this.eventType = eventType
-    this.version = version
+    this.eventOrder = eventOrder
     this.payload = payload
     this.sep = this.constructor.sep
   }
 
   get key () {
-    return `${this.timestamp}${this.sep}${this.eventId}${this.sep}${this.version}`
+    return `${this.timestamp}${this.sep}${this.eventId}${this.sep}${this.eventOrder}`
   }
 
   get value () {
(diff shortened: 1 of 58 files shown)


dannypaz:
[Fix orderbooks in a bad state](https://trello.com/c/mHyzzFJu/621-fix-orderbooks-in-a-bad-state)

broker-daemon/models/market-event.js
```diff
@@ -8,12 +8,13 @@ class MarketEvent {
...
     this.timestamp = timestamp
     this.eventType = eventType
+    this.eventOrder = eventOrder
     this.payload = payload
     this.sep = this.constructor.sep
   }
 
   get key () {
-    return `${this.timestamp}${this.sep}${this.eventId}`
+    return `${this.timestamp}${this.sep}${this.eventId}${this.sep}${this.eventOrder}`
```
martineehrlich: Maybe we should put this.eventOrder after timestamp so it provides ordering value.

broker-daemon/models/market-event.js
```diff
@@ -8,12 +8,13 @@ class MarketEvent {
     this.orderId = orderId
     this.timestamp = timestamp
     this.eventType = eventType
+    this.eventOrder = eventOrder
```
martineehrlich: I don't think we should use eventOrder as the name for this. It is confusing just because we use order and eventOrder in other places. Maybe ordering, sequence etc.

dannypaz: @martineehrlich Going to rename this value to `sequence` instead

broker-daemon/orderbook/index.js
```diff
@@ -112,26 +113,42 @@ class Orderbook {
...
+   * @returns {Object} res
+   * @returns {String} [lastUpdated=0] - nanosecond timestamp
+   * @returns {String} [sequence=0] - event version for a given timestamp
+   */
+  async lastUpdate () {
+    this.logger.info(`Retrieving last update from store for ${this.marketName}`)
 
-    this.logger.info(`Found last update of ${timestamp}`)
+    const {
+      timestamp: lastUpdated = 0,
```
martineehrlich: Shouldn't these 0's be strings?

PR #251, opened 2018-09-21
FIX - CLI common units to Broker base units

dannypaz:
## Description
This PR removes quantum unit calculation from the CLI and puts conversion from common to base units as the responsiblity of the Broker.

#### More Info:
Users will want to interact with the API in the way that is easiest and most familiar. That means using common units (e.g. BTC) instead of quantums (e.g. Satoshis).

The Broker Daemon should be responsible for converting from common units to quantums for interaction with the Relayer, Lightning nodes, etc.

#### Changes:
1. removed hardcoding of column size on the cli for `wallet balance`
2. removed quantum configuration from cli
3. removed quantum calculation from cli
4. added method on broker daemon `getWalletBalance` to convert common units to base units

## Related PRs

## Todos
- [ ] Tests
- [x] Documentation
- [x]  Link to Trello


dannypaz:
[All amounts and prices should be in decimals in the Broker API](https://trello.com/c/NRdaEFVA/725-all-amounts-and-prices-should-be-in-decimals-in-the-broker-api)

PR #252, opened 2018-09-24
FIX - Interchain router error handling during translation of swap

dannypaz:
## Description
If the lnd calls in `translateSwap` fail (for any reason), the error is not handled correctly by the interchain router. The result is an active htlc that destroys Node A's outbound channel.

This PR wraps `translateSwap` in a try-catch and throws a permanent error if the call itself fails, outside of error handling from lnd-engine.

## Related PRs

## Todos
- [x] Tests
- [x] Documentation
- [x] Link to Trello


broker-daemon/interchain-router/get-preimage.js
```diff
@@ -108,16 +107,21 @@ async function getPreimage ({ params, send, onCancel, onError, ordersByHash, eng
...
   }
 
   logger.debug(`Sending payment to ${takerAddress} to translate swap ${swapHash}`)
-  const { paymentPreimage, permanentError } = await outboundEngine.translateSwap(takerAddress, swapHash, outboundAmount, timeLockDelta)
+
+  try {
+    var { paymentPreimage, permanentError } = await outboundEngine.translateSwap(takerAddress, swapHash, outboundAmount, timeLockDelta)
+  } catch (err) {
+    logger.error(err)
+    return send({ permanentError: err.message })
```
treygriffith: you can only handle this as a permanent error if you can be certain that the error is permanent. Is that true in this case?

dannypaz: We cant guarantee it. Im going to make a change in lnd-engine for the situation I found. The error we want to handle is when no routes are found... however it looks like the call throws an error instead of returning an empty array for `routes`

PR #257, opened 2018-10-03
post route should not have id as param

martineehrlich:
## Description
For the createOrder rest route, we should not be specifying an id because no id exists on creation. 

## Related PRs
https://github.com/sparkswap/ccxt/pull/6

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Link to Trello


martineehrlich:
[add createOrder to ccxt implementation](https://trello.com/c/E5mhmieY/757-add-createorder-to-ccxt-implementation)

dannypaz:
@martineehrlich LGTM

PR #258, opened 2018-10-04
FIX - adjust get trades endpoint

dannypaz:
## Description
This PR makes a few small changes to the getTrades endpoint to allow the endpoint to function consistently with other RPCs. We've added functionality that was specific to ccxt, that is now removed in this PR including:

1. removes `info` from the getTrades rpc
2. removes millisecond timestamps from a trades response and instead opts to use nanoseconds

CCXT will handle `info` and the millisecond timestamps on its own as we expect the daemon to work in nanoseconds

## Related PRs

## Todos
- [x] Tests
- [x] Documentation
- [x] Link to Trello


dannypaz:
[add fetchTrades to ccxt implementation](https://trello.com/c/q4GRmqBP/756-add-fetchtrades-to-ccxt-implementation)

broker-daemon/models/market-event.js
```diff
@@ -63,24 +65,30 @@ class MarketEvent {
     const baseCommonAmount = Big(this.payload.baseAmount).div(baseCurrencyConfig.quantumsPerCommon)
     const counterCommonAmount = Big(this.payload.counterAmount).div(counterCurrencyConfig.quantumsPerCommon)
 
+    console.log(this.payload)
```
martineehrlich: logs

broker-daemon/models/market-event.js
```diff
@@ -63,24 +65,30 @@ class MarketEvent {
...
+    console.log(this.payload)
+
+    if (baseCommonAmount.eq(0)) {
+      return Big(0).toString()
```
martineehrlich: Did you figure out this issue?

dannypaz: This was only a dev line because of an out-of-date orderbook. We are all good

PR #259, opened 2018-10-04
Fix/add timestamps block

dannypaz:
## Description
This PR adds BlockOrder (order) creation timestamps to give the user an idea of when a blockorder was created on the broker.

This timestamp was added due to information needed in the CCXT `fetchOrder` command, however this is a good addition to our API for other users as well.

## Related PRs

## Todos
- [x] Tests
- [x] Documentation
- [x] Link to Trello

### DIFF
--- a/broker-cli/proto/broker.proto
+++ b/broker-cli/proto/broker.proto
@@ -296,8 +296,10 @@ message BlockOrder {
   TimeInForce time_in_force = 5;
   // Current status of the block order.
   BlockOrderStatus status = 6;
-  // Timestamp in nano-seconds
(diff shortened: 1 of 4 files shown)


dannypaz:
[add fetchOrder to ccxt implementation](https://trello.com/c/rAZwn24X/759-add-fetchorder-to-ccxt-implementation)

dannypaz:
@martineehrlich it is set on its own, but you can also specify a timestamp (which we use when we need to re-create BlockOrders). See: https://github.com/sparkswap/broker/pull/259/files#diff-6149506167c1b50a1ee6bb28a20b83afR26

martineehrlich:
Ah I see

PR #262, opened 2018-10-11
use crypto library instead of generate-safe-id

martineehrlich:
## Description
The generate-safe-id library we were using for unique ids of mongo records had a security vulnerability, so this PR switches to use crypto instead.

## Related PRs
https://github.com/sparkswap/relayer/pull/166

## Todos
- [x] Tests
- [x] Documentation
- [x] Link to Trello


martineehrlich:
[update `generate-safe-id` to a less vulnerable library](https://trello.com/c/MEkmktn4/580-update-generate-safe-id-to-a-less-vulnerable-library)

PR #263, opened 2018-10-12
remove disable relayer auth flag as relayer auth should always be enabled

martineehrlich:
## Description
We should remove the flag where we can disable auth from the relayer, as this was only used for development and should now be required for dev going forward.

## Todos
- [x] Tests
- [x] Documentation
- [x] Link to Trello


martineehrlich:
[Remove `DISABLE_RELAYER_AUTH` from broker](https://trello.com/c/SlHgPIWc/790-remove-disablerelayerauth-from-broker)

PR #267, opened 2018-10-29
Feature: add optional cors support for the http server

treygriffith:
## Description
As part of the Lightning Hackday, I made a web GUI for viewing the orderbook. To do so, I needed to enable CORS for the Broker to allow the web GUI to pull the data.

Rather than letting that interesting functionality go away, I added it as an option to the http server. There currently is not a user-configurable way to turn it on, but that can be added later fairly easily by adding some configuration to the environment and Broker Daemon constructor.

## Todos
- [x] Tests
- [ ] Documentation
- [ ] Link to Trello


treygriffith:
To take our offline conversation online:
This does not expose us any further than grpc already does. In the future, we should specify the domains that are allowed, especially before exposing this to configuration.

PR #269, opened 2018-10-31
Cannot cancel order

martineehrlich:
## Description
There is a bug when cancelling orders that was introduced when the BlockOrder.fromStore method was implemented. The blockorder.fromstore in the cancel function needs to be awaited so that it is the blockOrder object and not a promise.

## Todos
- [x] Tests
- [x] Documentation
- [x] Link to Trello


martineehrlich:
[Cannot cancel orderbook orders](https://trello.com/c/Zj0WK9iT/835-cannot-cancel-orderbook-orders)

PR #270, opened 2018-11-01
Prevent user from trying to fill own order

martineehrlich:
## Description
If a user places a bid, and then attempts to fill an order, we should make sure that we do not try to have the user fill their own order. We should throw an error if the order the user is trying to fill is their own.

In order to complete this ticket, I:
1) Added a sublevel index for orderId on the orders store so we could look up an order by orderId (also did an investigation on switching the primary key from blockOrderId to orderId, which is accomplishable but should be done as a separate piece of work). 
2) When attempting to fillOrders, check all the bestOrders and see if they are in the orders database, we can assume that the order belongs to the user (the ordersStore is only the users own orders)
3) Throw an error if the orderId of order going to be filled is in the list of the users orderIds.

## Todos
- [x] Tests
- [x] Documentation
- [x] Link to Trello

### DIFF
--- a/broker-daemon/block-order-worker/index.js
+++ b/broker-daemon/block-order-worker/index.js
@@ -38,6 +38,9 @@ class BlockOrderWorker extends EventEmitter {
     const filterOrdersWithHash = (key, value) => !!Order.fromStorage(key, value).swapHash
     const getHashFromOrder = (key, value) => Order.fromStorage(key, value).swapHash
 
+    const filterOrdersWithOrderId = (key, value) => !!Order.fromStorage(key, value).orderId
+    const getOrderIdFromOrder = (key, value) => Order.fromStorage(key, value).orderId
+
     // create an index for the ordersStore so that orders can be retrieved by their swapHash
     this.ordersByHash = new SublevelIndex(
       this.ordersStore,
@@ -47,6 +50,16 @@ class BlockOrderWorker extends EventEmitter {
       // only index orders that have a swap hash defined
       filterOrdersWithHash
     )
+
+    // create an index for the ordersStore so that orders can be retrieved by their orderId
+    this.ordersByOrderId = new SublevelIndex(
+      this.ordersStore,
+      'ordersByOrderId',
+      // index by orderId
+      getOrderIdFromOrder,
+      // only index orders that have an orderId defined
+      filterOrdersWithOrderId
+    )
   }
 
   /**
@@ -55,6 +68,7 @@ class BlockOrderWorker extends EventEmitter {
    */
   async initialize () {
     await this.ordersByHash.ensureIndex()
+    await this.ordersByOrderId.ensureIndex()
   }
 
   /**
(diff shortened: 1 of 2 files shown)


broker-daemon/block-order-worker/index.js
```diff
@@ -402,55 +416,69 @@ class BlockOrderWorker extends EventEmitter {
...
-    const promisedFills = orders.map(async (order) => {
+    const ordersByOrderId = await Promise.all(orders.map((order) => {
+      const range = {
+        gte: order.orderId,
+        lte: order.orderId
+      }
+      return getRecords(this.ordersByOrderId, Order.fromStorage.bind(Order), this.ordersByOrderId.range(range))
+    }))
+
+    const ownOrderIds = ordersByOrderId.filter(order => order && order.length > 0).map(([order]) => order.orderId)
```
treygriffith: confusing to use `order` here for an array of orders, better to use something like `orders`

broker-daemon/block-order-worker/index.js
```diff
@@ -402,55 +416,69 @@ class BlockOrderWorker extends EventEmitter {
       throw new Error(`No engine available for ${counterSymbol}`)
     }
 
-    const promisedFills = orders.map(async (order) => {
+    const ordersByOrderId = await Promise.all(orders.map((order) => {
```
treygriffith: this is a bit of a confusing name for what this variable holds, especially since it's the same name as the index

broker-daemon/block-order-worker/index.js
```diff
@@ -402,55 +416,69 @@ class BlockOrderWorker extends EventEmitter {
...
+        gte: order.orderId,
+        lte: order.orderId
+      }
+      return getRecords(this.ordersByOrderId, Order.fromStorage.bind(Order), this.ordersByOrderId.range(range))
+    }))
+
+    const ownOrderIds = ordersByOrderId.filter(order => order && order.length > 0).map(([order]) => order.orderId)
+
+    const promisedFills = []
+    for (let order of orders) {
```
treygriffith: why convert this to a for loop from a `map`?

broker-daemon/block-order-worker/index.js
```diff
@@ -402,55 +416,69 @@ class BlockOrderWorker extends EventEmitter {
...
+          {
+            relayer,
+            engines,
+            logger,
+            store
+          },
+          blockOrder.id,
+          order,
+          { fillAmount }
+        ).then((fsm) => {
```
treygriffith: is it worth it to move this function somewhere else? Like a dedicated `applyFsmListeners` function? Feels a bit messy in the loop

broker-daemon/block-order-worker/index.spec.js
```diff
@@ -140,12 +140,15 @@ describe('BlockOrderWorker', () => {
       expect(worker).to.have.property('engines', engines)
     })
 
-    it('creates an index for the orders store', async () => {
-      expect(SublevelIndex).to.have.been.calledOnce()
+    it('creates indices for the orders store by swaphash and orderId', async () => {
```
treygriffith: I would separate these tests (although I know the `calledTwice` thing is a bit awkward)

broker-daemon/block-order-worker/index.js
```diff
@@ -402,55 +416,69 @@ class BlockOrderWorker extends EventEmitter {
...
+    for (let order of orders) {
       const depthRemaining = targetDepth.minus(currentDepth)
 
       // if we have already reached our target depth, create no further fills
       if (depthRemaining.lte(0)) {
         return
       }
 
+      if (ownOrderIds.includes(order.orderId)) {
+        throw new Error(`Cannot fill own order ${order.orderId}`)
```
treygriffith: this creates an interesting edge case: if the first order that you want to fill is another user's, but the second is your own, it will fill the first but error out fo the second. Should we handle that separately as part of more general error handling, or should we tackle it now?

treygriffith: I think we should leave it for now, but open a ticket specifically outlining this case. It's part of the whole "more nuanced approach to error handling" theme.

martineehrlich: https://trello.com/c/I08iGste/844-better-error-handling-retrying-when-filling-own-orders

dannypaz:
@treygriffith fine w/ me. @martineehrlich LGTM

PR #271, opened 2018-11-01
Feature: Retry orderbook when watchMarket fails

treygriffith:
## Description
This PR allows the Broker to stay up when `watchMarket` fails, and retry `watchMarket` until it comes back up. It throws an error if trying to use the orderbook which relies on being synced with the Relayer.

## Todos
- [x] Tests
- [x] Documentation

### DIFF
--- a/broker-daemon/orderbook/index.js
+++ b/broker-daemon/orderbook/index.js
@@ -60,7 +60,7 @@ class Orderbook {
    * @returns {Promise<Array<MarketEventOrder>>} A promise that resolves an array of MarketEventOrder records
    */
   async all () {
-    this._assertSynced()
+    this.assertSynced()
     this.logger.info(`Retrieving all records for ${this.marketName}`)
     return getRecords(this.store, MarketEventOrder.fromStorage.bind(MarketEventOrder))
   }
@@ -73,7 +73,7 @@ class Orderbook {
    * @return {Array<Object>} trades
    */
   async getTrades (since, limit) {
-    this._assertSynced()
+    this.assertSynced()
     const params = {limit}
     if (since) {
       const sinceDate = new Date(since).toISOString()
@@ -97,7 +97,7 @@ class Orderbook {
    * @return {Promise<BestOrders>} A promise that resolves MarketEventOrders of the best priced orders
    */
   getBestOrders ({ side, depth, quantumPrice }) {
-    this._assertSynced()
+    this.assertSynced()
     return new Promise((resolve, reject) => {
       this.logger.info(`Retrieving best priced from ${side} up to ${depth}`)
 
@@ -151,7 +151,7 @@ class Orderbook {
    * @returns {Array<MarketEventOrder>}
    */
   async getOrderbookEventsByTimestamp (timestamp) {
-    this._assertSynced()
+    this.assertSynced()
     return getRecords(
       this.store,
       (key, value) => JSON.parse(value),
@@ -167,7 +167,7 @@ class Orderbook {
    * @returns {Array<MarketEventOrder>}
    */
   async getMarketEventsByTimestamp (timestamp) {
-    this._assertSynced()
+    this.assertSynced()
     return getRecords(
       this.eventStore,
       (key, value) => JSON.parse(value),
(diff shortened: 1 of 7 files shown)


broker-daemon/orderbook/index.js
```diff
@@ -130,13 +144,56 @@ class Orderbook {
...
+      MarketEvent.rangeFromTimestamp(timestamp)
+    )
+  }
+
+  /**
+   * Ensures that the orderbook is synced before accessing it
+   * @return {void}
+   * @throws {Error} If Orderbook is not synced to Relayer
+   */
+  _assertSynced () {
```
dannypaz: [nit] Is this a convention we want to use going forward? I remember having a convo about this a few months ago and we decided to omit `underscores`. If we wanted a truly private method, we could add the function to the top of this file, and not export it.

broker-daemon/relayer/market-watcher.js
```diff
@@ -0,0 +1,126 @@
...
+class MarketWatcher extends EventEmitter {
+  /**
+   * Set up the watcher
+   * @param  {grpc.ServerStreaming} watcher        Stream of events from the Relayer
+   * @param  {Sublevel}             store          Leveldb compatible store
+   * @param  {Object}               RESPONSE_TYPES Response types from the proto file
+   * @param  {Object} logger
+   * @return {MarketWatcher}
+   */
+  constructor (watcher, store, RESPONSE_TYPES, logger) {
```
dannypaz: should we just call this `stream` instead of watcher?

broker-daemon/orderbook/index.js
```diff
@@ -28,19 +28,30 @@ class Orderbook {
...
+      this.bidIndex = await (new BidIndex(this.store)).ensureIndex()
 
-    return this.logger.info(`Market ${this.marketName} initialized.`)
+      this.synced = true
+      this.logger.info(`Market ${this.marketName} initialized.`)
+    })
+
+    watcher.on('end', (error) => {
+      this.logger.info(`Market ${this.marketName} unavailable, retrying in 5s`, { error })
+      setTimeout(() => {
```
dannypaz: TODO for the exponential backoff stuff from @martineehrlich

broker-daemon/orderbook/index.js
```diff
@@ -150,7 +211,7 @@ class Orderbook {
 
   /**
    * Gets the last time this market was updated with data from the relayer
-   *
+   * @private
```
martineehrlich: lol

broker-daemon/relayer/market-watcher.js
```diff
@@ -0,0 +1,132 @@
...
+    super()
+    this.watcher = watcher
+    this.store = store
+    this.migrating = null
+    this.logger = logger
+    this.RESPONSE_TYPES = RESPONSE_TYPES
```
martineehrlich: I am confused as to why you are passing this in, when we have hard coded expected response types lower in the file?

treygriffith: They aren't hardcoded lower in the file, we're pulling them out of the proto file in relayer-client.js

dannypaz: @treygriffith when we upgrade the grpc on broker, we will no longer have access to those enums from the protofile. As a preventative measure, I would hardcode them somewhere

broker-daemon/orderbook/index.js
```diff
@@ -8,6 +8,8 @@ const nano = require('nano-seconds')
 const consoleLogger = console
 consoleLogger.debug = console.log.bind(console)
 
+const RETRY_WATCHMARKET = 5000
```
dannypaz: Add docs

broker-daemon/relayer/market-watcher.js
dannypaz: @treygriffith Wondering if there is a more descriptive term for this object `watcher`

broker-daemon/relayer/market-watcher.js
dannypaz: NVM, we can hold off until the actual migration. per our talk @treygriffith

PR #272, opened 2018-11-02
Retry lnd engine validation

martineehrlich:
## Description
If LND is not ready, sparkswapd will fail to start. It will retry 10 times with PM2, but frequently LND needs much longer (e.g. re-indexing the blockchain).

1) Add exponentialBackoff
2) Make validating engines a background process

## Related PRs

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Link to Trello

### DIFF
--- a/broker-daemon/index.js
+++ b/broker-daemon/index.js
@@ -8,7 +8,7 @@ const Orderbook = require('./orderbook')
 const BlockOrderWorker = require('./block-order-worker')
 const BrokerRPCServer = require('./broker-rpc/broker-rpc-server')
 const InterchainRouter = require('./interchain-router')
-const { logger } = require('./utils')
+const { logger, exponentialBackoff } = require('./utils')
 const CONFIG = require('./config')
 
 /**
@@ -170,17 +170,10 @@ class BrokerDaemon {
           this.logger.info(`Initializing BlockOrderWorker`)
           await this.blockOrderWorker.initialize()
           this.logger.info('BlockOrderWorker initialized')
-        })(),
-        ...Array.from(this.engines, async ([ symbol, engine ]) => {
-          this.logger.info(`Validating engine configuration for ${symbol}`)
-          try {
-            await engine.validateNodeConfig()
-          } catch (e) {
-            throw new Error(`Engine for ${symbol} failed to validate: ${e.message || e.details}`)
-          }
-          this.logger.info(`Validated engine configuration for ${symbol}`)
-        })
+        })()
       ])
+
+      this.validateEngines()
 
       this.rpcServer.listen(this.rpcAddress)
       this.logger.info(`BrokerDaemon RPC server started: gRPC Server listening on ${this.rpcAddress}`)
@@ -233,6 +226,18 @@ class BrokerDaemon {
 
     this.orderbooks.set(marketName, new Orderbook(marketName, this.relayer, this.store.sublevel(marketName), this.logger))
     return this.orderbooks.get(marketName).initialize()
+  }
+
+  async validateEngines () {
+    await Promise.all(Array.from(this.engines, async ([ symbol, engine ]) => {
+      try {
+        this.logger.info(`Validating engine configuration for ${symbol}`)
+        await engine.validateNodeConfig()
+        this.logger.info(`Validated engine configuration for ${symbol}`)
+      } catch (e) {
+        await exponentialBackoff(async () => engine.validateNodeConfig(), 10, 1000, () => console.log(`Validated engine configuration for ${symbol}`))
+      }
+    }))
   }
 }
 
--- a/broker-daemon/index.js
+++ b/broker-daemon/index.js
@@ -251,12 +251,12 @@ class BrokerDaemon {
    * We do not await this function because we want the validations to run in the background.
    * It can take time for the engines to be ready, so we use exponential backoff to retry validation
    * for a period of time, until it is either successful or there is actually something wrong.
-   * @returns {<void>}
+   * @returns {void}
    */
   async validateEngines () {
     this.engines.forEach(async (engine, symbol) => {
       try {
-        await exponentialBackoff(() => { engine.validateNodeConfig() }, EXPONENTIAL_BACKOFF_ATTEMPTS, EXPONENTIAL_BACKOFF_DELAY, {symbol})
+        await exponentialBackoff(() => { return engine.validateNodeConfig() }, EXPONENTIAL_BACKOFF_ATTEMPTS, EXPONENTIAL_BACKOFF_DELAY, {symbol})
       } catch (e) {
         this.logger.error(`Failed to validate engine for ${symbol}, error: ${e}`)
         return
--- a/broker-daemon/index.js
+++ b/broker-daemon/index.js
@@ -44,7 +44,7 @@ const DEFAULT_INTERCHAIN_ROUTER_ADDRESS = '0.0.0.0:40369'
  * @constant
  * @type {String}
  */
-const EXPONENTIAL_BACKOFF_ATTEMPTS = 10
+const EXPONENTIAL_BACKOFF_ATTEMPTS = 24
 
 /**
  * Delay in each retry attempt to validating an engine
@@ -52,7 +52,7 @@ const EXPONENTIAL_BACKOFF_ATTEMPTS = 10
  * @constant
  * @type {Integer} milliseconds
  */
-const EXPONENTIAL_BACKOFF_DELAY = 1000
+const EXPONENTIAL_BACKOFF_DELAY = 5000
 
 /**
  * Default host and port that the Relayer is set up on
--- a/broker-daemon/index.js
+++ b/broker-daemon/index.js
@@ -245,11 +245,17 @@ class BrokerDaemon {
   }
 
   async validateEngines () {
-    await Promise.all(Array.from(this.engines, async ([ symbol, engine ]) => {
-      this.logger.info(`Validating engine configuration for ${symbol}`)
-      exponentialBackoff(() => { return engine.validateNodeConfig() }, EXPONENTIAL_BACKOFF_ATTEMPTS, EXPONENTIAL_BACKOFF_DELAY)
+    this.engines.forEach(async (engine, symbol) => {
+      try {
+        this.logger.info(`Validating engine configuration for ${symbol}`)
+        await exponentialBackoff(() => { engine.validateNodeConfig() }, EXPONENTIAL_BACKOFF_ATTEMPTS, EXPONENTIAL_BACKOFF_DELAY)
+      } catch (e) {
+        this.logger.error(`Failed to validate engine for ${symbol}, error: ${e}`)
+        return
+      }
+
       this.logger.info(`Validated engine configuration for ${symbol}`)
-    }))
+    })
   }
 }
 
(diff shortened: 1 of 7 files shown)


broker-daemon/utils/exponential-backoff.js
```diff
@@ -0,0 +1,21 @@
+async function exponentialBackoff (callFunction, attempts, delay, callback) {
```
dannypaz: jsdoc information

broker-daemon/index.js
```diff
@@ -234,6 +227,18 @@ class BrokerDaemon {
...
   }
+
+  async validateEngines () {
+    await Promise.all(Array.from(this.engines, async ([ symbol, engine ]) => {
+      try {
+        this.logger.info(`Validating engine configuration for ${symbol}`)
+        await engine.validateNodeConfig()
+        this.logger.info(`Validated engine configuration for ${symbol}`)
+      } catch (e) {
+        await exponentialBackoff(async () => engine.validateNodeConfig(), 10, 1000, () => console.log(`Validated engine configuration for ${symbol}`))
```
dannypaz: remove the anonymous functions in this method signature and instead create them in `exponentialBackoff`

Example: `await exponentialBackoff(engine.validateNodeConfig, 10, 1000,...`

dannypaz: @martineehrlich actually, we should be able to remove that async anonymous function altogether? `engine.validateNodeConfig()` is already async

broker-daemon/utils/exponential-backoff.js
```diff
@@ -0,0 +1,21 @@
...
+  try {
+    var res = await callFunction()
+  } catch (e) {
+    console.log(e)
```
dannypaz: we should use logger here `logger.error`

broker-daemon/utils/exponential-backoff.js
```diff
@@ -0,0 +1,21 @@
...
+  }
+
+  if (res) {
+    callback(res)
+  } else {
+    if (attempts > 0) {
+      setTimeout(function () {
```
dannypaz: There are timer limits, but im not sure if we would ever hit them (its like 24 days or something). With that said, do we want to limit the amount of attempts?

broker-daemon/index.js
dannypaz: [nit] we might want to add constants for `10` and `1000` w/ documentation on the values?

```
EXPONENTIAL_BACKOFF_ATTEMPTS = 10
EXPONENTIAL_BACKOFF_DELAY = 1000
```

broker-daemon/utils/exponential-backoff.js
```diff
@@ -0,0 +1,19 @@
...
+const delay = require('./delay')
+
+async function exponentialBackoff (callFunction, attempts, delayTime) {
+  try {
+    var res = callFunction()
+  } catch (error) {
+    logger.error(`Error (${error}) with ${callFunction}, retry attempts left: ${attempts}`)
+    if (attempts > 0) {
+      await delay(delayTime)
+      exponentialBackoff(callFunction, --attempts, delayTime * 2)
```
treygriffith: need to assign the return value here for `res`, otherwise you won't get the return value

broker-daemon/utils/exponential-backoff.js
```diff
@@ -0,0 +1,19 @@
...
+    } else {
+      logger.error(`Error (${error}) with ${callFunction}, no retry attempts left`)
```
treygriffith: This should re-throw since it has fully failed

broker-daemon/utils/exponential-backoff.js
```diff
@@ -0,0 +1,19 @@
+const logger = require('./logger')
+const delay = require('./delay')
+
+async function exponentialBackoff (callFunction, attempts, delayTime) {
+  try {
+    var res = callFunction()
+  } catch (error) {
+    logger.error(`Error (${error}) with ${callFunction}, retry attempts left: ${attempts}`)
```
treygriffith: might be better to put the error as a trailing parameter

broker-daemon/index.js
```diff
@@ -234,6 +243,20 @@ class BrokerDaemon {
     this.orderbooks.set(marketName, new Orderbook(marketName, this.relayer, this.store.sublevel(marketName), this.logger))
     return this.orderbooks.get(marketName).initialize()
   }
+
+  async validateEngines () {
```
treygriffith: add some docs, especially w/r/t how you are not supposed to `await` it

broker-daemon/index.js
```diff
@@ -170,18 +186,11 @@ class BrokerDaemon {
...
-            await engine.validateNodeConfig()
-          } catch (e) {
-            throw new Error(`Engine for ${symbol} failed to validate: ${e.message || e.details}`)
-          }
-          this.logger.info(`Validated engine configuration for ${symbol}`)
-        })
+        })()
       ])
 
+      this.validateEngines()
```
treygriffith: I would make a note here that this will run in the background until the engines are validated, otherwise it's a bit weird to have a side-effects only call like this

broker-daemon/utils/exponential-backoff.js
```diff
@@ -0,0 +1,19 @@
+const logger = require('./logger')
+const delay = require('./delay')
+
+async function exponentialBackoff (callFunction, attempts, delayTime) {
```
treygriffith: add docs to this

treygriffith: add some tests for this

broker-daemon/index.js
treygriffith: add some tests for this

broker-daemon/utils/exponential-backoff.spec.js
```diff
@@ -0,0 +1,49 @@
...
+    expect(result).to.eql(res)
+  })
+
+  it('retries the callFunction if there was an error and there are still retries left', async () => {
+    callFunction = sinon.stub().rejects('Error')
+    try {
+      await exponentialBackoff(callFunction, attempts, delayTime, logOptions)
+      expect(logger.error).to.have.been.called()
+      expect(delay).to.have.been.calledWith(delayTime)
+      expect(callFunction).to.have.been.calledThrice()
```
treygriffith: It might be worth it to check:
1. the call order (that callFunction gets called after delay)
2. that callFunction only gets called after delay resolves

broker-daemon/index.js
```diff
@@ -234,6 +245,26 @@ class BrokerDaemon {
...
+   * It can take time for the engines to be ready, so we use exponential backoff to retry validation
+   * for a period of time, until it is either successful or there is actually something wrong.
+   * @returns {<void>}
+   */
+  async validateEngines () {
+    this.engines.forEach(async (engine, symbol) => {
+      try {
+        await exponentialBackoff(() => { engine.validateNodeConfig() }, EXPONENTIAL_BACKOFF_ATTEMPTS, EXPONENTIAL_BACKOFF_DELAY, {symbol})
+      } catch (e) {
+        this.logger.error(`Failed to validate engine for ${symbol}, error: ${e}`)
```
treygriffith: Do we just want to error here? Or do we need to take some other action (like crash the process)?

broker-daemon/index.spec.js
```diff
@@ -414,6 +406,60 @@ describe('broker daemon', () => {
...
+      expect(btcEngine.validateNodeConfig).to.have.been.called()
+
+      const validateLtcEngine = exponentialBackoff.args[1][0]
+      validateLtcEngine()
+
+      expect(ltcEngine.validateNodeConfig).to.have.been.called()
+
+      expect(exponentialBackoff).to.have.been.calledTwice()
+      expect(exponentialBackoff).to.have.been.calledWith(
+        sinon.match.func,
```
treygriffith: you should add some tests for the function that gets passed to `exponentialBackoff` to make sure that it does what you expect it to (i.e. that it calls `validateNodeConfig`)

broker-daemon/utils/delay.spec.js
```diff
@@ -0,0 +1,19 @@
...
+
+  beforeEach(() => {
+    setTimeout = sinon.stub().callsArg(0)
+    delayTime = 100
+    delay.__set__('setTimeout', setTimeout)
+  })
+
+  it('calls the call function', async () => {
+    await delay(delayTime)
+    expect(setTimeout).to.have.been.called()
```
treygriffith: check that it gets called with a function and the milliseconds you expect

treygriffith: check that delay won't resolve until setTimeout calls the first arg

broker-daemon/utils/exponential-backoff.js
```diff
@@ -0,0 +1,36 @@
...
+ * @return {Promise}
+ */
+async function exponentialBackoff (callFunction, attempts, delayTime, logOptions = {}) {
+  try {
+    var res = await callFunction()
+  } catch (error) {
+    if (attempts > 0) {
+      logger.error(`Error calling ${callFunction}. Retrying in ${delayTime / 1000} seconds, attempts left: ${attempts - 1}`, logOptions)
+      await delay(delayTime)
+      res = await exponentialBackoff(callFunction, --attempts, delayTime * DELAY_MULTIPLIER)
```
treygriffith: is it worth it to have a new variable that is `attempts - 1` since you use it in the log and in the next call to exponentialBackoff ?

broker-daemon/index.spec.js
treygriffith: I see that you actually did this - maybe just split it up into different tests so that it is clear?

martineehrlich:
[Wait for LND to be ready before trying to start sparkswapd](https://trello.com/c/cKzRhlyM/608-wait-for-lnd-to-be-ready-before-trying-to-start-sparkswapd)

broker-daemon/index.js
```diff
@@ -234,6 +245,26 @@ class BrokerDaemon {
     this.orderbooks.set(marketName, new Orderbook(marketName, this.relayer, this.store.sublevel(marketName), this.logger))
     return this.orderbooks.get(marketName).initialize()
   }
+
+  /**
+   * Validates engines
+   * We do not await this function because we want the validations to run in the background.
+   * It can take time for the engines to be ready, so we use exponential backoff to retry validation
+   * for a period of time, until it is either successful or there is actually something wrong.
+   * @returns {<void>}
```
treygriffith: [nit] no need for the angle brackets here (I don't think)

broker-daemon/index.js
```diff
@@ -234,6 +245,26 @@ class BrokerDaemon {
...
+   */
+  async validateEngines () {
+    this.engines.forEach(async (engine, symbol) => {
+      try {
+        await exponentialBackoff(() => { engine.validateNodeConfig() }, EXPONENTIAL_BACKOFF_ATTEMPTS, EXPONENTIAL_BACKOFF_DELAY, {symbol})
```
treygriffith: `engine.validateNodeConfig` needs to return, otherwise `exponentialBackoff` won't wait for it to complete

broker-daemon/index.spec.js
```diff
@@ -414,6 +406,60 @@ describe('broker daemon', () => {
...
+    it('passes engine validation to exponential backoff for each engine', () => {
+      brokerDaemon.validateEngines()
+
+      const validateBtcEngine = exponentialBackoff.args[0][0]
+      validateBtcEngine()
+
+      expect(btcEngine.validateNodeConfig).to.have.been.called()
+
+      const validateLtcEngine = exponentialBackoff.args[1][0]
+      validateLtcEngine()
```
treygriffith: check that the return value of this is a promise

PR #273, opened 2018-11-05
Fix: url encode order id format to make it compatible with the CLI

treygriffith:
## Description
This change applies base64 url encoding to block order ids when they are being generated so that they are compatible with the existing CLI rules and so that they are effectively future-proofed for web-based clients.

### DIFF
--- a/broker-daemon/utils/generate-id.js
+++ b/broker-daemon/utils/generate-id.js
@@ -16,9 +16,9 @@ function generateId () {
  */
 function urlEncode (base64Str) {
   return base64Str
-    .replace(/=/g, "")
-    .replace(/\+/g, "-")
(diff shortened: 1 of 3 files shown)


broker-daemon/utils/generate-id.js
```diff
@@ -2,7 +2,14 @@ const crypto = require('crypto')
 
 function generateId () {
   const data = crypto.randomBytes(20).toString('hex')
-  return crypto.createHash('sha256').update(data).digest('base64')
+  return urlEncode(crypto.createHash('sha256').update(data).digest('base64'))
+}
+
+function urlEncode (base64Str) {
```
dannypaz: add jsdoc to this and `generateId`

PR #276, opened 2018-11-06
Feature: partial fill re-order

treygriffith:
## Description
This PR automatically re-places orders when a previous order was filled but not completely filled.

## Todos
- [x] Tests
- [x] Documentation

### DIFF
--- a/broker-daemon/block-order-worker/index.js
+++ b/broker-daemon/block-order-worker/index.js
@@ -346,7 +346,7 @@ class BlockOrderWorker extends EventEmitter {
     })
 
     // if the fill for this order isn't for the entire order, re-place the remainder
-    osm.once('execute', async () => {
+    osm.once('before:execute', async () => {
       this.logger.debug(`Order ${osm.order.orderId} has been filled, re-placing the remainder`)
       try {
         const remainingBaseAmount = Big(osm.order.baseAmount).minus(osm.order.fillAmount)
--- a/broker-daemon/block-order-worker/index.js
+++ b/broker-daemon/block-order-worker/index.js
@@ -300,7 +300,7 @@ class BlockOrderWorker extends EventEmitter {
 
     const blockOrder = await this.getBlockOrder(blockOrderId)
 
-    // check the fillamount on each collection of state machines from the block order
+    // check the fillAmount on each collection of state machines from the block order
     // and make sure that either is equal to how much we are trying to fill.
     let totalFilled = Big(0)
     totalFilled = blockOrder.fills.reduce((acc, fsm) => acc.plus(fsm.fill.fillAmount), totalFilled)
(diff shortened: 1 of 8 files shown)


broker-daemon/state-machines/plugins/state-machine-events.js
```diff
@@ -17,14 +17,25 @@ class StateMachineEvents extends StateMachinePlugin {
...
     const plugin = this
 
     return {
       /**
+       * Emit an event before a transition (prefixed by 'before:')
+       * e.g. stateMachine.on('before:create')
+       * @param  {Object} lifecycle
+       * @return {void}
+       */
+      onBeforeTransition: function (lifecycle) {
```
dannypaz: @treygriffith did you want to add this in this PR? It doesn't look like we use it anywhere

broker-daemon/block-order-worker/index.js
```diff
@@ -326,6 +327,53 @@ class BlockOrderWorker extends EventEmitter {
...
+    // if the fill for this order isn't for the entire order, re-place the remainder
+    osm.once('execute', async () => {
+      this.logger.debug(`Order ${osm.order.orderId} has been filled, re-placing the remainder`)
+      try {
+        const remainingBaseAmount = Big(osm.order.baseAmount).minus(osm.order.fillAmount)
+        // if there is no remaining base amount (i.e. the entire order was filled)
+        // take no action as the block order completion is handled in the `on('complete')` listener
+        if (remainingBaseAmount.gt(0)) {
+          this.logger.debug(`Re-placing an order for ${remainingBaseAmount.toString()} for Block Order ${blockOrder.id}`)
+          await this.workBlockOrder(blockOrder, remainingBaseAmount)
```
dannypaz: @treygriffith Im a little confused. Shouldn't `workBlockOrder` be handling this? Or are we saying that `workBlockOrder` will set up the initial orders, but if we get to execute and something has failed, we will need to rerun `workBlockOrder`?

broker-daemon/state-machines/plugins/state-machine-events.js
treygriffith: @dannypaz I just pushed a change that does use it - I accidentally left it out, thanks for the catch.

broker-daemon/block-order-worker/index.js
treygriffith: Before this change, `workBlockOrder` (or more specifically, `workLimitBlockOrder` and `workMarketBlockOrder`) determined the depth to fill by looking at the total depth of the block order.

With this change, those methods now take a parameter to specify the limit so that we can do things like re-place orders with less the initial amount of the block order.

The other alternative is to fetch all the orders associated with this block order from the store and make a holistic determination about how much is left on the order. That seemed inefficient and unnecessary to me, which is why I did it this way.

dannypaz: The was I understand it is that `workBlockOrder` will try to fill as much as possible and then just return. Why would we not add the placement of an order in `workBlockOrder` instead of on the `execute` action?

treygriffith: To take our offline discussion online: you are thinking of when initially working the order making sure that you first fill any orders at the given price before placing the remainder. That already happens (in `workLimitBlockOrder`), this is handling the scenario where a counterparty fills part, but not all, of your own placed order.

PR #277, opened 2018-11-07
Fix: watchmarket leak

treygriffith:
## Description
My last PR (#271) introduced an issue where index hooks were getting re-applied to the store multiple times, resulting in errors when trying to delete keys.

This change fixes that in two ways:
1. By removing index hooks as part of the clearIndex operation (for general hygiene)
2. Having indexes only be built on startup, and re-use the hooks for handling when the orderbook needs to be started from scratch, since it uses regular `del` operations anyway.

## Related PRs
#271 Introduced the issue.

## Todos
- [x] Tests
- [x] Documentation

### DIFF
--- a/broker-daemon/orderbook/index.js
+++ b/broker-daemon/orderbook/index.js
@@ -34,6 +34,8 @@ class Orderbook {
     this.eventStore = store.sublevel('events')
     this.index = new OrderbookIndex(store, this.eventStore, this.marketName)
     this.store = this.index.store
+    this.askIndex = new AskIndex(this.store)
+    this.bidIndex = new BidIndex(this.store)
     this.logger = logger
     this.synced = false
   }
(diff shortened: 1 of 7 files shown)


broker-daemon/orderbook/orderbook-index.js
```diff
@@ -64,7 +68,7 @@ class OrderbookIndex {
    * Create a hook for new events added to the store to modify the orderbook
    */
   _addIndexHook () {
-    this.eventStore.pre((dbOperation, add) => {
+    this._removeHook = this.eventStore.pre((dbOperation, add) => {
```
dannypaz: naming is confusing. was not expecting `pre` to return this function. Per our talk, would totally appreciate having this code be more clear

broker-daemon/relayer/market-watcher.js
```diff
@@ -64,6 +64,8 @@ class MarketWatcher extends EventEmitter {
   async handleResponse (response) {
     const { RESPONSE_TYPES } = this
 
+    // TODO: I don't think this will work well if there is more than one event queued since they all
```
dannypaz: Can you explain `why` in more detail. Do you think this would cause us to get into a weird state? What would be the situation?

broker-daemon/orderbook/index.js
```diff
@@ -49,35 +51,45 @@ class Orderbook {
...
-
       this.synced = true
-      this.logger.info(`Market ${this.marketName} initialized.`)
+      this.logger.info(`Market ${this.marketName} synced.`)
     })
 
     watcher.on('end', (error) => {
       this.logger.info(`Market ${this.marketName} unavailable, retrying in ${RETRY_WATCHMARKET}ms`, { error })
 
       // TODO: exponential backoff?
```
dannypaz: I added a card for this to make sure we didnt forget to add it once https://github.com/sparkswap/broker/pull/272 is in

PR #279, opened 2018-11-10
Feature: Orderbook Checksum

treygriffith:
## Description
This PR adds a checksum to the orderbook representing its current state. It compares that checksum against the checksums sent by the Relayer, and if there is a mismatch, it dumps its current state in order to re-sync.

This allows the Broker to detect when it is out of sync with the Relayer (due to a dropped event) and get back into sync.

## Todos
- [x] Tests
- [x] Documentation

## Future considerations
This does not take advantage of (or validate) that indexing that we do to create the orderbook itself (i.e. the `OrderbookIndex`) primarily because I wanted to keep the checksum localized to processing new Relayer events. However, it may become a performance issue at some point, or be a useful way to validate the Orderbook Index itself against the checksum.

### DIFF
--- a/broker-daemon/orderbook/index.js
+++ b/broker-daemon/orderbook/index.js
@@ -68,29 +68,44 @@ class Orderbook {
   /**
    * Sync orderbook state with the Relayer and retry when it fails
    * @private
-   * @return {Promise} Resolves when market is being watched (not necessarily when it is synced)
+   * @param {Object}   options
+   * @param {Boolean}  options.flush Whether or not to flush our current copy of the orderbook or try to update it
+   * @return {Promise}               Resolves when market is being watched (not necessarily when it is synced)
    */
-  async watchMarket () {
+  async watchMarket ({ flush = false } = {}) {
     this.logger.debug(`Watching market ${this.marketName}...`)
 
     const { baseSymbol, counterSymbol } = this
-    const { lastUpdated, sequence } = await this.lastUpdate()
-    const params = { baseSymbol, counterSymbol, lastUpdated, sequence }
+    const params = { baseSymbol, counterSymbol }
+
+    // If we are re-using our current orderbook, find the last event we know about
+    // and use that to request future events.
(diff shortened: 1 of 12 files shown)


broker-daemon/orderbook/index.js
```diff
@@ -68,7 +68,7 @@ class Orderbook {
   /**
    * Sync orderbook state with the Relayer and retry when it fails
    * @private
-   * @return {Promise} Resolves when market is being watched (not necessarily when it is synced)
+   * @return {Promise}               Resolves when market is being watched (not necessarily when it is synced)
```
dannypaz: [nit] not sure if you meant to add this huge space here

broker-daemon/orderbook/index.spec.js
```diff
@@ -258,11 +260,23 @@ describe('Orderbook', () => {
       expect(relayer.watchMarket).to.have.been.calledWith(eventStore, sinon.match({ baseSymbol, counterSymbol, lastUpdated, sequence }))
     })
 
+    it('sets a sync listener', () => {
+      expect(watcher.once).to.have.been.calledWith('sync', sinon.match.func)
+    })
+
+    xit('sets an end listener', () => {
```
dannypaz: Not sure if you meant to have this `xit` here

broker-daemon/relayer/market-watcher.js
```diff
@@ -26,6 +27,17 @@ class MarketWatcher extends EventEmitter {
     this.RESPONSE_TYPES = RESPONSE_TYPES
 
     this.setupListeners()
+    this.createChecksum()
```
martineehrlich: I think you should set `this.checksum = new Checksum()` here instead of in createChecksum. As it is right now, it's a bit confusing to set an instance variable in an obscure function instead of here.

You could have a function that populates the checksum that you call after instantiating it here.

Similar thoughts for this.migratating and this.migrate()

PR #281, opened 2018-11-13
Fix - Engine validation changes

dannypaz:
## Description
In the past, we've used a synchronous way of validating a broker's engines so that we would prevent the broker from starting until all nodes were available.

However, we have now changed the validations of engines to occur async to allow the broker to be operational even if a single node is not available.

With this said, the validation of an engine now occurs on [engine initialization](https://github.com/sparkswap/lnd-engine/pull/110) and is the responsibility of the engine itself.

## Related PRs
#110 

## Todos
- [x] Tests
- [x] Documentation
- [x] Link to Trello

### DIFF
--- a/broker-daemon/index.js
+++ b/broker-daemon/index.js
@@ -8,7 +8,7 @@ const Orderbook = require('./orderbook')
 const BlockOrderWorker = require('./block-order-worker')
 const BrokerRPCServer = require('./broker-rpc/broker-rpc-server')
 const InterchainRouter = require('./interchain-router')
-const { logger } = require('./utils')
+const { logger, exponentialBackoff } = require('./utils')
 const CONFIG = require('./config')
 
 /**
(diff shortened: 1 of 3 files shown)


broker-daemon/index.js
```diff
@@ -253,17 +253,8 @@ class BrokerDaemon {
...
-      try {
-        await exponentialBackoff(() => { return engine.validateNodeConfig() }, EXPONENTIAL_BACKOFF_ATTEMPTS, EXPONENTIAL_BACKOFF_DELAY, {symbol})
-      } catch (e) {
-        this.logger.error(`Failed to validate engine for ${symbol}, error: ${e}`)
-        return
-      }
-
-      this.logger.info(`Validated engine configuration for ${symbol}`)
-    })
+  validateEngines () {
```
treygriffith: do we need to validate engines separately, or can we do it while we are creating the engines?

dannypaz:
[Move Engine validation and retry into Engine](https://trello.com/c/EfceFCwG/877-move-engine-validation-and-retry-into-engine)

broker-daemon/index.js
dannypaz: @treygriffith having it separated is easier to read. When we assign `engines` we create them and map them to `this.engines` and it felt weird to validate inside of the map.

PR #282, opened 2018-11-14
Feature/retry filling relayer cancelled orders

martineehrlich:
## Description
When attempting to fill an order, we look in the database for the best matching orders and then attempt to fill them. There is a high probability that by the time we attempt to fill an order someone else has already filled it. If this is the case, we want to rework the block order to find the best orders again (non filled) and retry. 

I added a metadata flag in the linked relayer PR, that gets passed back to the broker if the order is not in a placed state. If the FSM gets a reject event when creating the fill or filling the order, and the metadata is passed back, we want to work the block order again. 

## Related PRs
https://github.com/sparkswap/relayer/pull/184

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Link to Trello


broker-daemon/state-machines/plugins/events.js
```diff
@@ -31,7 +34,7 @@ class StateMachineEvents extends StateMachinePlugin {
        * @return {void}
        */
       onBeforeTransition: function (lifecycle) {
-        plugin.eventHandler.emit(`before:${lifecycle.transition}`)
+        this.eventHandler.emit(`before:${lifecycle.transition}`)
```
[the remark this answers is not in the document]
martineehrlich: Ah yeah, I'll add a comment. Thanks!

broker-daemon/block-order-worker/index.js
```diff
@@ -524,11 +524,14 @@ class BlockOrderWorker extends EventEmitter {
     // We are hooking into the reject lifecycle event of a fill state machine to trigger
     // the failure of a blockorder
     fsm.once('reject', () => {
-      this.failBlockOrder(blockOrder.id, fsm.fill.error)
-        .catch(e => {
+      fsm.removeAllListeners()
+      if (fsm.fill.error.metadata && fsm.fill.error.metadata.get('invalidOrderStatus')) {
```
dannypaz: [nit] would move `invalidOrderStatus` to a constant

broker-daemon/block-order-worker/index.spec.js
```diff
@@ -1234,25 +1234,71 @@ describe('BlockOrderWorker', () => {
...
 
-      it('fails a block order if the call is rejected', async () => {
+      it('reworks a blockOrder if the order was already filled', async () => {
+        await worker._fillOrders(blockOrder, orders, targetDepth)
+        worker.workBlockOrder = sinon.stub().resolves(true)
+        await onceStub.args[1][1]()
+        expect(worker.workBlockOrder).to.have.been.calledWith(blockOrder, Big('100'))
+      })
+
+      it('fails a block order if the call is rejected and the order was not already filled', async () => {
```
treygriffith: I would call this something like `fails a block order if the call is rejected for reason other than the order is in the wrong state`

broker-daemon/block-order-worker/index.spec.js
```diff
@@ -1234,25 +1234,71 @@ describe('BlockOrderWorker', () => {
...
-        expect(onceStub).to.have.been.calledWith('reject', sinon.match.func)
+      it('removes all listeners if the call is rejected', async () => {
+        await worker._fillOrders(blockOrder, orders, targetDepth)
+        worker.workBlockOrder = sinon.stub().resolves(true)
+        await onceStub.args[1][1]()
+        expect(fill.removeAllListeners).to.have.been.calledOnce()
       })
 
-      it('fails a block order if the call is rejected', async () => {
+      it('reworks a blockOrder if the order was already filled', async () => {
```
treygriffith: I would call this `reworks a blockOrder if the order is not in a state to be filled` since it could be in another state for lots of reasons (cancelled, not enough funds in channel, etc)

broker-daemon/block-order-worker/index.js
```diff
@@ -528,14 +528,21 @@ class BlockOrderWorker extends EventEmitter {
...
+    /**
+     * We are hooking into the reject lifecycle event of a fill state machine to trigger the failure of a blockorder.
+     * If the error comes back with an ORDER_NOT_PLACED, it means that the order the fillStateMachine
+     * attempted to fill was not in a state to be filled and we should rework the blockOrder
+     */
     fsm.once('reject', () => {
-      this.failBlockOrder(blockOrder.id, fsm.fill.error)
-        .catch(e => {
+      fsm.removeAllListeners()
+      if (fsm.isRelayerError()) {
```
dannypaz: `isRelayerError` adds logic from the fsm into this reject handler, where instead its better to use `fsm.shouldRetry` or something like that

broker-daemon/state-machines/fill-state-machine.js
```diff
@@ -313,6 +334,13 @@ const FillStateMachine = StateMachine.factory({
     onBeforeReject: function (lifecycle, error) {
       this.logger.error(`Encountered error during transition, rejecting`, error)
       this.fill.error = error
+    },
+    /**
+     * Returns true if there is a relayer error associated with the fill, false if not
```
dannypaz: [nit] maybe make a note that this method has no transition

broker-daemon/state-machines/fill-state-machine.js
```diff
@@ -313,6 +334,13 @@ const FillStateMachine = StateMachine.factory({
...
+     * @return {Boolean}
+     */
+    isRelayerError: function () {
```
dannypaz: were you able to use `get isRelayerError() {}` here or did it error on you?

broker-daemon/state-machines/fill-state-machine.js
```diff
@@ -313,6 +334,13 @@ const FillStateMachine = StateMachine.factory({
...
+      return !!this.fill.error && this.fill.error.code === FILL_ERROR_CODES.ORDER_NOT_PLACED
```
dannypaz: sorry for being a stickler about `FILL_ERROR_CODES` :-/

broker-daemon/state-machines/fill-state-machine.js
```diff
@@ -20,6 +20,16 @@ const {
  */
 const UNASSIGNED_PREFIX = 'NO_ASSIGNED_ID_'
 
+/**
+ * Error codes that can come back from relayer
+ * @type {Object}
+ * @constant
+ * @default
+ */
+const FILL_ERROR_CODES = {
```
dannypaz: use `Object.freeze()` for this constant

broker-daemon/state-machines/fill-state-machine.js
martineehrlich: errored

broker-daemon/state-machines/fill-state-machine.spec.js
```diff
@@ -626,6 +650,33 @@ describe('FillStateMachine', () => {
     })
   })
 
+  describe('#shouldRetry', () => {
+    let fakeFill
+    let fsm
+
+    beforeEach(() => {
+      fsm = new FillStateMachine({ store, logger, relayer, engines })
+      fakeFill = { error: { message: 'ORDER_NOT_PLACED' } }
```
dannypaz: [nit] to make it easier on us if this string changes, we can use the constant value instead of hardcode. If you decide to make this change, you'll have changes in this file and `order-state-machine.spec.js` too

```
const { ORDER_NOT_PLACED } = FillStateMachine.__get__('FILL_ERROR_CODES')
fakeFill = { error: { message: 'ORDER_NOT_PLACED' } }
```

PR #283, opened 2018-11-15
Reject block orders where the user does not have enough funds committed

martineehrlich:
## Description
Rather than creating a block order and having the individual orders get rejected, the Broker Daemon should reject block order creation where there are not enough funds to cover the orders desired.

In createBlockOrder, we now check that the amount of the order is less than or equal to the amount the user has in that channel

## Todos
- [x] Tests
- [x] Documentation
- [x] Link to Trello

### DIFF
--- a/broker-daemon/block-order-worker/index.js
+++ b/broker-daemon/block-order-worker/index.js
@@ -102,18 +102,20 @@ class BlockOrderWorker extends EventEmitter {
     const blockOrder = new BlockOrder({ id, marketName, side, amount, price, timeInForce })
     const baseEngine = this.engines.get(blockOrder.baseSymbol)
     const counterEngine = this.engines.get(blockOrder.counterSymbol)
-    const { address: counterSymbolAddress } = await this.relayer.paymentChannelNetworkService.getAddress({symbol: blockOrder.counterSymbol})
-    const { address: baseSymbolAddress } = await this.relayer.paymentChannelNetworkService.getAddress({symbol: blockOrder.baseSymbol})
+    const [{address: counterSymbolAddress}, {address: baseSymbolAddress}] = await Promise.all([
+      this.relayer.paymentChannelNetworkService.getAddress({symbol: blockOrder.counterSymbol}),
+      this.relayer.paymentChannelNetworkService.getAddress({symbol: blockOrder.baseSymbol})
+    ])
 
     if (blockOrder.isBid) {
-      const outboundBalanceIsSufficient = await counterEngine.isBalanceSufficient(counterSymbolAddress, Big(blockOrder.counterAmount))
+      const outboundBalanceIsSufficient = await counterEngine.isBalanceSufficient(counterSymbolAddress, blockOrder.counterAmount)
 
       // If the user tries to place an order for more than they hold in the counter engine channel, throw an error
       if (!outboundBalanceIsSufficient) {
         throw new Error(`Insufficient funds in outbound ${blockOrder.counterSymbol} channel to create order`)
       }
 
-      const inboundBalanceIsSufficient = await baseEngine.isBalanceSufficient(baseSymbolAddress, Big(blockOrder.baseAmount), {outbound: false})
+      const inboundBalanceIsSufficient = await baseEngine.isBalanceSufficient(baseSymbolAddress, blockOrder.baseAmount, {outbound: false})
       // If the user tries to place an order and the relayer does not have the funds to complete in the base channel, throw an error
       if (!inboundBalanceIsSufficient) {
         throw new Error(`Insufficient funds in inbound ${blockOrder.baseSymbol} channel to create order`)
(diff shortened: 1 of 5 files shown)


martineehrlich:
[Reject block orders where the user does not have enough funds committed](https://trello.com/c/GRBUgpNz/291-reject-block-orders-where-the-user-does-not-have-enough-funds-committed)

broker-daemon/block-order-worker/index.js
```diff
@@ -101,6 +101,24 @@ class BlockOrderWorker extends EventEmitter {
 
     const blockOrder = new BlockOrder({ id, marketName, side, amount, price, timeInForce })
 
+    // If the user tries to place an order for more than they hold in the counter engine channel, throw an error
+    if (blockOrder.side === BlockOrder.SIDES['BID']) {
```
treygriffith: [nit] `BlockOrder.SIDES.BID` would make more sense to me than using a string here

broker-daemon/block-order-worker/index.js
```diff
@@ -101,6 +101,24 @@ class BlockOrderWorker extends EventEmitter {
...
+      const counterEngine = this.engines.get(blockOrder.counterSymbol)
```
treygriffith: it's probably worth checking that the counter engine exists

broker-daemon/block-order-worker/index.js
```diff
@@ -101,6 +101,24 @@ class BlockOrderWorker extends EventEmitter {
...
+      const amountInChannel = await counterEngine.getTotalChannelBalance()
```
treygriffith: don't we have an engine method that specifically checks if there are sufficient funds?

broker-daemon/block-order-worker/index.js
dannypaz: @treygriffith @martineehrlich Not sure if this in the scope of this PR, but this is probably a case for setting an `isBid` property on the BlockOrder, so the check would be `if (blockOrder.isBid())`.

dannypaz: If we agree on creating the property, we should move the line at 114 too

broker-daemon/block-order-worker/index.js
martineehrlich: It is checked above? I didn't think it was necessary to check again.

broker-daemon/block-order-worker/index.js
martineehrlich: I'm gunna go for it! I think it's a good idea 👍

broker-daemon/block-order-worker/index.js
```diff
@@ -100,6 +100,41 @@ class BlockOrderWorker extends EventEmitter {
     }
 
     const blockOrder = new BlockOrder({ id, marketName, side, amount, price, timeInForce })
+    const baseEngine = this.engines.get(blockOrder.baseSymbol)
+    const counterEngine = this.engines.get(blockOrder.counterSymbol)
+    const { address: counterSymbolAddress } = await this.relayer.paymentChannelNetworkService.getAddress({symbol: blockOrder.counterSymbol})
+    const { address: baseSymbolAddress } = await this.relayer.paymentChannelNetworkService.getAddress({symbol: blockOrder.baseSymbol})
```
treygriffith: I would run this in parallel to the other call to the payment channel network service

broker-daemon/block-order-worker/index.js
```diff
@@ -100,6 +100,41 @@ class BlockOrderWorker extends EventEmitter {
...
+
+    if (blockOrder.isBid) {
+      const outboundBalanceIsSufficient = await counterEngine.isBalanceSufficient(counterSymbolAddress, Big(blockOrder.counterAmount))
```
treygriffith: does the `isBalanceSufficient` take `Big` as the type for it's amount? I don't think it should, and if it doesn't, I would not do this conversion here.

broker-daemon/block-order-worker/index.js
```diff
@@ -100,6 +100,41 @@ class BlockOrderWorker extends EventEmitter {
     }
 
     const blockOrder = new BlockOrder({ id, marketName, side, amount, price, timeInForce })
+    const baseEngine = this.engines.get(blockOrder.baseSymbol)
+    const counterEngine = this.engines.get(blockOrder.counterSymbol)
+    const { address: counterSymbolAddress } = await this.relayer.paymentChannelNetworkService.getAddress({symbol: blockOrder.counterSymbol})
+    const { address: baseSymbolAddress } = await this.relayer.paymentChannelNetworkService.getAddress({symbol: blockOrder.baseSymbol})
+
+    if (blockOrder.isBid) {
```
treygriffith: Have you considered ways to consolidate this with the `isAsk` block below (without sacrificing readability)?

martineehrlich: I would prefer to clean this up in the next PR when I combine with the earmarking. Right now the code is very readable and I would like to leave it but I will condense in the next PR if thats cool with you.

broker-daemon/models/block-order.js
```diff
@@ -172,6 +172,22 @@ class BlockOrder {
     })
   }
 
+  /**
+   * get boolean for if the blockOrder is a bid
```
dannypaz: [nit] @martineehrlich we can omit docs for these `get` properties since they are self documenting

PR #286, opened 2018-11-20
Feature: Turn off fees and deposits

treygriffith:
## Description
This PR adds support for the new Relayer flag to disable fees and deposits.

It refactors invoice payment into a utility that is used by the OrderStateMachine and FillStateMachine and respects the new flag.

I have yet to write tests, but I wanted to get feedback on the approach and choices that I made before I spend a bunch of time unit testing.

## Todos
- [x] Tests
- [x] Documentation


treygriffith:
[Turn fees and deposits off](https://trello.com/c/Zx8LN5Nl/857-turn-fees-and-deposits-off)

broker-daemon/utils/pay-invoice.js
```diff
@@ -0,0 +1,25 @@
+async function payInvoice (required, paymentRequest, engine, logger, type, publicId) {
```
martineehrlich: jsdoc

broker-daemon/models/order.js
```diff
@@ -153,11 +157,67 @@ class Order {
...
+      throw new Error(`paramsForPlace: orderId is missing.`)
+    }
+
+    if (!outboundSymbol) {
+      throw new Error(`paramsForPlace: outboundSymbol is missing.`)
+    }
+
+    return {
+      feePaymentRequest,
+      feeRequired,
```
dannypaz: passing these fields to the relayer on place or create feels weird. The relayer dictates if the fees/deposits are required and shouldn't have to rely on the broker for this information. Am I misunderstanding this?

treygriffith: We don't pass that to the Relayer. I should probably change the docs for this.

We use that flag in the OrderStateMachine to determine whether or not we pay the fee, but we don't send that to the Relayer.

broker-daemon/state-machines/fill-state-machine.js
```diff
@@ -200,34 +200,25 @@ const FillStateMachine = StateMachine.factory({
...
+        throw new Error(`No engine available for ${outboundSymbol}`)
       }
 
-      const [feeRefundPaymentRequest, depositRefundPaymentRequest] = await Promise.all([
-        outboundEngine.createRefundInvoice(feePaymentRequest),
-        outboundEngine.createRefundInvoice(depositPaymentRequest),
-        outboundEngine.payInvoice(feePaymentRequest),
-        outboundEngine.payInvoice(depositPaymentRequest)
+      const [ feeRefundPaymentRequest, depositRefundPaymentRequest ] = await Promise.all([
+        payInvoice(feeRequired, feePaymentRequest, outboundEngine, this.logger, 'fee', fillId),
```
dannypaz: we should constantize these types (or use the invoice types from the model)

broker-daemon/state-machines/fill-state-machine.js
```diff
@@ -214,9 +214,29 @@ const FillStateMachine = StateMachine.factory({
...
+
+      if (depositRequired) {
+        payDepositInvoice = payInvoice(outboundEngine, depositPaymentRequest)
+      } else {
+        this.logger.debug(`Skipping paying deposit invoice for ${fillId}, not required`)
+      }
+
+      const [
+        feeRefundPaymentRequest,
+        depositRefundPaymentRequest
```
dannypaz: This is a question for below, but will the call to `fillOrder` fail if these are undefined (meaning they are not provided in the request to the relayer)?

broker-daemon/utils/pay-invoice.js
```diff
@@ -8,28 +8,37 @@
...
-  })
-
-  logger.info(`Paid ${type} for ${publicId}`)
-
   return refundPaymentRequest
 }
 
 module.exports = payInvoice
+
+// async function payInvoice (engine, paymentRequest) {
```
dannypaz: forgot to remove the comments

broker-daemon/state-machines/fill-state-machine.js
treygriffith: fillOrder won't fail if these are undefined (so long as the fee/deposit was not required)

PR #288, opened 2018-11-20
Fix: Reset orderbook when resetting the event store

treygriffith:
## Description
The change which caused the event store to flush when encountering an invalid checksum (#279) does not flush the `OrderbookIndex` itself as the `OrderbookIndex` is designed only to handle `put` events, and ignores `del` events.

This created an issue where stale orders would show up if they existed in the `OrderbookIndex` before the market event store was flushed.

Ideally, the `OrderbookIndex` could handle `del` events (much like the `BidIndex` and `AskIndex` already do) but since that would require some serious changes, instead this change adds a call to reset the `OrderbookIndex` when migrating the event store.

This is an instance of the decision made in #279 to only validate the checksum when processing new events coming back to bite us. We should re-examine in the future if we want to do another checksum within the `OrderbookIndex` to avoid issues like this.

## Related PRs
#279 

## Todos
- [x] Tests
- [x] Documentation

### DIFF
--- a/broker-daemon/orderbook/index.js
+++ b/broker-daemon/orderbook/index.js
@@ -118,6 +118,7 @@ class Orderbook {
 
       this.logger.info(`Market ${this.marketName} encountered sync'ing error, re-building`, { error })
       await watcher.migrate()
+      await this.index.ensureIndex()
       this.watchMarket()
     }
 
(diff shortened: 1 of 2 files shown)


dannypaz:
@treygriffith this looks good to me. Whether or not we want to add additional documentation (specifically to the line where we need to `ensureIndex` is up to you.

PR #292, opened 2018-11-26
add helper functions for inbound/outbound

martineehrlich:
## Description
We need these helper functions as a part of earmarking funds. 

## Related PRs

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Link to Trello

### DIFF
--- a/broker-daemon/models/block-order.js
+++ b/broker-daemon/models/block-order.js
@@ -130,6 +130,38 @@ class BlockOrder {
   }
 
   /**
+  * Convenience getter for outboundAmount
+  * @return {String} String representation of the amount of currency we will send outbound for the order
+  */
+  get outboundAmount () {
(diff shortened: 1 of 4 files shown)


broker-daemon/models/block-order.js
```diff
@@ -129,6 +129,38 @@ class BlockOrder {
...
+  /**
+  * Convenience getter for inboundAmount
+  * @return {String} String representation of the amount of currency we will receive inbound for the order
+  */
+  get inboundAmount () {
+    return this.isBid ? this.baseAmount : this.counterAmount
+  }
+
+  /**
+  * Get the symbol of the currency we will receive inbound if the order is completed
```
dannypaz: @martineehrlich we should remove `if the order is completed` to the docs for these functions

dannypaz: Remove the text in `order.js` too

PR #294, opened 2018-11-27
Feature/unlock status

dannypaz:
## Description
Housekeeping PR for `healthcheck` to add correct representations of the state of an engine/relayer

screen shot 2018-11-27 at 3 03 33 pm

screen shot 2018-11-27 at 3 05 48 pm

## Related PRs
In relation to the unlock change here https://github.com/sparkswap/lnd-engine/pull/117

## Todos
- [x] Tests
- [x] Documentation
- [x] Link to Trello


dannypaz:
[[EPIC] Add status check to let the user know they need to unlock wallet](https://trello.com/c/iuBFXBuz/860-epic-add-status-check-to-let-the-user-know-they-need-to-unlock-wallet)

broker-cli/broker-daemon-client/index.js
```diff
@@ -66,7 +66,7 @@ class BrokerDaemonClient {
 
     if (this.disableAuth) {
       // TODO: Eventually allow broker daemon client to use the cli's logger
-      console.warn('disableAuth is set to true. The CLI will try to connect to the daemon without ssl')
+      console.warn('`disableAuth` is set to true. The CLI will try to connect to the daemon without SSL/TLS'.yellow)
```
treygriffith: it's kind of weird to use `colors` without requiring them in this file

broker-cli/utils/error-handler.js
```diff
@@ -1,13 +1,11 @@
...
 /**
  * Takes in an error object and throws a friendly error if the broker daemon is down
  *
  * @param {Error} error
  * @return {void}
  */
-
 function handleError (error) {
-  if (error.message === '14 UNAVAILABLE: Connect Failed') {
+  if (error.message && error.message === '14 UNAVAILABLE: Connect Failed') {
```
treygriffith: seems like an unnecessary addition. Did you mean to do `if (error && error.message ==...)`?

broker-daemon/broker-rpc/admin-service/health-check.js
```diff
@@ -8,27 +8,32 @@ const STATUS_CODES = Object.freeze({
...
+ * @param {Array} req
+ * @param {String} req.symbol
+ * @param {Object} req.engine
+ * @return {Object} res
+ * @return {String} res.symbol - an engine's currency symbol (e.g. BTC or LTC)
+ * @return {String} res.status - an engine's status
  */
-async function getEngineStatus (engine) {
+async function getEngineStatus ([ symbol, engine ]) {
+  let status = STATUS_CODES.OK
```
treygriffith: it would make the most sense to me to have `status` start off as `UNKNOWN` and only change to `OK` after checking that the engine `isAvailable`. I know the result is the same, but it looks weird this way.

broker-cli/utils/error-handler.js
dannypaz: Nope, this was just a brain fart. thanks for the catch

broker-cli/broker-daemon-client/index.js
dannypaz: I like the use of color here (because grey text isn't as interesting of a warning). ill add colors to the top JIC

broker-daemon/broker-rpc/admin-service/health-check.js
dannypaz: I might omit the default entirely just because an `UNKNOWN` status would never get hit with the current code

PR #295, opened 2018-11-29
FEATURE - Add force release for channels

dannypaz:
## Description
This PR adds a `force` flag to the `sparkswap wallet release` command that allows the user to force close all channels that are open w/ the exchange.

The Broker may run into an issue where it has open channels with the relayer, however the relayer may be down or offline. We want the broker to be able to close those channels HOWEVER the only way possible is by forcing the closing of the channel through the engines.

if the call to `sparkswap wallet release` fails, we will prompt the user to the user the `--force` flag if they understand that force releasing could take an extended amount of time AND this may add additional fees to the closing tx of the channel.

## Related PRs

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Link to Trello

### DIFF
--- a/broker-cli/commands/wallet.js
+++ b/broker-cli/commands/wallet.js
@@ -547,7 +547,7 @@ module.exports = (program) => {
     .option('--market [marketName]', 'Relevant market name', validations.isMarketName)
     .command(`wallet ${SUPPORTED_COMMANDS.RELEASE}`, 'Closes channels open on the specified market')
     .option('--market <marketName>', 'Relevant market name', validations.isMarketName, null, true)
-    .option('--force [force]', 'Relevant market name', null, false)
+    .option('--force', 'Relevant market name', null, false)
     .command(`wallet ${SUPPORTED_COMMANDS.WITHDRAW}`, 'Withdraws specified amount of coin from wallet')
     .argument('<symbol>', `Supported currencies: ${SUPPORTED_SYMBOLS.join('/')}`)
     .argument('<amount>', 'Amount of currency to commit to the relayer', validations.isDecimal)
--- a/broker-cli/commands/wallet.js
+++ b/broker-cli/commands/wallet.js
@@ -300,16 +300,37 @@ function formatBalance (balance, status) {
  * @return {Void}
  */
 async function release (args, opts, logger) {
-  const { rpcAddress, market } = opts
+  const { market } = args
+  const { rpcAddress, force } = opts
 
   try {
     const client = new BrokerDaemonClient(rpcAddress)
 
-    const answer = await askQuestion(`Are you sure you want to close all channels you have open on the ${market} market? (Y/N)`)
+    let question
+    if (force) {
+      question = `Are you sure you want to FORCE the release of all channels you have open on the ${market} market? (Y/N)`
+    } else {
+      question = `Are you sure you want to release all channels you have open on the ${market} market? (Y/N)`
+    }
+
+    const answer = await askQuestion(question)
 
     if (!ACCEPTED_ANSWERS.includes(answer.toLowerCase())) return
 
-    await client.walletService.releaseChannels({market})
+    try {
+      await client.walletService.releaseChannels({ market, force })
+    } catch (e) {
+      logger.info(`Failed to release payment channels for ${market}`.red)
+
+      if (!force) {
+        logger.info('You can force the release of funds using the `--force` option, however')
+        logger.info('this has the potential to lock your funds for an extended period of time')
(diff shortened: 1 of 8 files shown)


dannypaz:
[Add force option to smart release](https://trello.com/c/znlWaEau/651-add-force-option-to-smart-release)

broker-cli/commands/wallet.js
```diff
@@ -300,16 +300,37 @@ function formatBalance (balance, status) {
...
+
+    const answer = await askQuestion(question)
 
     if (!ACCEPTED_ANSWERS.includes(answer.toLowerCase())) return
 
-    await client.walletService.releaseChannels({market})
+    try {
+      await client.walletService.releaseChannels({ market, force })
+    } catch (e) {
+      logger.error(`Failed to release payment channels for ${market}`.red)
```
martineehrlich: Do we want to show what the error is before we advise them to force close?

treygriffith: 👍

broker-cli/commands/wallet.js
```diff
@@ -521,7 +546,8 @@ module.exports = (program) => {
     .command(`wallet ${SUPPORTED_COMMANDS.NETWORK_STATUS}`, 'Payment Channel Network status for trading in different markets')
     .option('--market [marketName]', 'Relevant market name', validations.isMarketName)
     .command(`wallet ${SUPPORTED_COMMANDS.RELEASE}`, 'Closes channels open on the specified market')
-    .option('--market [marketName]', 'Relevant market name', validations.isMarketName)
+    .option('--market <marketName>', 'Relevant market name', validations.isMarketName, null, true)
+    .option('--force [force]', 'Relevant market name', null, false)
```
treygriffith: I don't think force takes a parameter, so it should just be `--force`, not `--force [force]`, right? Unless I'm misunderstanding the caporal syntax

broker-cli/proto/broker.proto
```diff
@@ -921,6 +921,8 @@ message CommitRequest {
 message ReleaseChannelsRequest {
   // Market to close channels on
   string market = 1;
+  // Force channels to close which may result in funds being locked for additional time and fees
```
treygriffith: Add a note that it includes inactive and pending channels

broker-cli/commands/wallet.js
dannypaz: @treygriffith @martineehrlich the error gets rethrown after the text.

broker-cli/commands/wallet.js
dannypaz: nope! you are all good. thanks for the catch

broker-cli/commands/wallet.js
martineehrlich: Yes, but I was suggesting that we should show/log what the error is here, and still throw after.

broker-cli/commands/wallet.js
```diff
@@ -300,16 +300,37 @@ function formatBalance (balance, status) {
...
     if (!ACCEPTED_ANSWERS.includes(answer.toLowerCase())) return
 
-    await client.walletService.releaseChannels({market})
+    try {
+      await client.walletService.releaseChannels({ market, force })
+    } catch (e) {
+      logger.error(`Failed to release payment channels for ${market}`.red)
+
+      if (!force) {
+        logger.info('You can force the release of funds using the `--force` option, however')
```
treygriffith: [nit] maybe use a multi-line string here

broker-cli/commands/wallet.js
```diff
@@ -521,7 +546,8 @@ module.exports = (program) => {
     .command(`wallet ${SUPPORTED_COMMANDS.NETWORK_STATUS}`, 'Payment Channel Network status for trading in different markets')
     .option('--market [marketName]', 'Relevant market name', validations.isMarketName)
     .command(`wallet ${SUPPORTED_COMMANDS.RELEASE}`, 'Closes channels open on the specified market')
-    .option('--market [marketName]', 'Relevant market name', validations.isMarketName)
+    .option('--market <marketName>', 'Relevant market name', validations.isMarketName, null, true)
+    .option('--force', 'Relevant market name', null, false)
```
treygriffith: this description for `--force` looks wrong

dannypaz: @treygriffith it is sooo wrong. thanks for the catch

broker-cli/commands/wallet.js
dannypaz: @treygriffith the formatting was easier w/ individual lines

PR #296, opened 2018-11-29
remove cyclic dependencies and add madge to check for cycle dependencies

martineehrlich:
## Description
We had a cyclic dependency with blockOrder requiring orderStateMachine and Order and then orderStateMachine requiring Order from the index. We need Order to be required directly to avoid the dependency, which also brings up the question of whether or not we should even have index files as they make it more likely to introduce cycles. 

Also added the madge package to avoid this in the future.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Link to Trello

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,12 +15,14 @@
     "dist-cli": "npm run broker-proto && bash ./scripts/publish-broker-cli.sh",
     "format": "standard --fix",
     "pretest": "npm run format; exit 0",
-    "test": "NODE_PATH=. mocha 'broker-cli/{,!(node_modules)/**}/**/*.spec.js' 'broker-daemon/**/**/*.spec.js'",
+    "mocha-test": "NODE_PATH=. mocha 'broker-cli/{,!(node_modules)/**}/**/*.spec.js' 'broker-daemon/**/**/*.spec.js'",
+    "test": "npm run mocha-test && npm run check-for-cycles",
     "coverage": "nyc npm test",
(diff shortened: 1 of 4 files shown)


martineehrlich:
[Bug: cyclic dependency in OrderStateMachine and FillStateMachine](https://trello.com/c/PlzdotQk/932-bug-cyclic-dependency-in-orderstatemachine-and-fillstatemachine)

martineehrlich:
[Bug: cyclic dependency in OrderStateMachine and FillStateMachine](https://trello.com/c/PlzdotQk/932-bug-cyclic-dependency-in-orderstatemachine-and-fillstatemachine)

package.json
```diff
@@ -15,12 +15,14 @@
     "dist-cli": "npm run broker-proto && bash ./scripts/publish-broker-cli.sh",
     "format": "standard --fix",
     "pretest": "npm run format; exit 0",
-    "test": "NODE_PATH=. mocha 'broker-cli/{,!(node_modules)/**}/**/*.spec.js' 'broker-daemon/**/**/*.spec.js'",
+    "mocha-test": "NODE_PATH=. mocha 'broker-cli/{,!(node_modules)/**}/**/*.spec.js' 'broker-daemon/**/**/*.spec.js'",
+    "test": "npm run mocha-test && npm run check-for-cycles",
```
treygriffith: I think you should check for cycles before running the tests

PR #299, opened 2018-12-03
Cancel outstanding orders on failed block orders

martineehrlich:
## Description
We want to cancel all outstanding orders associated with a failed block order. Currently we do this for cancelled orders. We also should not try to work a block order if it is not in an active state.

- Cancel outstanding orders if we are failing the block orders
- Do not call failBlockOrder from cancelBlockOrder, just directly fail the blockOrder if cancelling the outstanding orders failed to avoid a cycle.
- Check the status of a blockOrder before working it

## Related PRs

## Todos
- [x] Tests
- [x] Documentation
- [x] Link to Trello

### DIFF
--- a/broker-daemon/block-order-worker/index.js
+++ b/broker-daemon/block-order-worker/index.js
@@ -175,6 +175,11 @@ class BlockOrderWorker extends EventEmitter {
     return blockOrder
   }
 
+  /**
+   * Cancels all outstanding orders for the given block order
+   * @param {BlockOrder}
+   * @return {Void}
+   */
   async cancelOutstandingOrders (blockOrder) {
     await blockOrder.populateOrders(this.ordersStore)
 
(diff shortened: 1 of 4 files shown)


martineehrlich:
[Fail block order should actually affect its state](https://trello.com/c/abLInw0s/874-fail-block-order-should-actually-affect-its-state)

broker-daemon/block-order-worker/index.js
```diff
@@ -258,6 +268,8 @@ class BlockOrderWorker extends EventEmitter {
   async workBlockOrder (blockOrder, targetDepth) {
     this.logger.info('Working block order', { blockOrderId: blockOrder.id })
 
+    if (!blockOrder.isInWorkableState) throw new Error('BlockOrder is not in a state to be worked')
```
treygriffith: I don't think we should throw here, for two reasons:

1. This is not an exception per se - if a block order places two orders, one of which gets filled and the other of which ends in an error and fails the block order, the one that gets filled will try to re-work the block order.
2. `workBlockOrder` only ever happens async, so this will just result in an unhandled rejection

PR #301, opened 2018-12-04
FIX - close channels on all available engines

dannypaz:
## Description
This PR updates the `sparkswap wallet release` command to close channels from all available engines AND additionally prompt the user if there had been errors when trying to release funds.

**Implementation Notes:**
I wanted this feature to be a simple as possible. Right now, we'll fail the call if engines/configuration is messed up and then we'll handle errors for each individual `closeChannel` call for the market. The user can check `network-status` to see what market/channels could potentially be open, but there may need to be an enhancement there.

Example:
screen shot 2018-12-06 at 3 06 23 pm

## Related PRs

## Todos
- [x] Tests
- [x] Documentation
- [x] Link to Trello

### DIFF
--- a/broker-cli/commands/wallet.js
+++ b/broker-cli/commands/wallet.js
@@ -318,20 +318,28 @@ async function release (args, opts, logger) {
     if (!ACCEPTED_ANSWERS.includes(answer.toLowerCase())) return
 
     try {
-      await client.walletService.releaseChannels({ market, force })
-    } catch (e) {
-      logger.error(`Failed to release payment channels for ${market}`.red)
+      const { errors } = await client.walletService.releaseChannels({ market, force })
 
-      if (!force) {
-        logger.info('You can force the release of funds using the `--force` option, however')
-        logger.info('this has the potential to lock your funds for an extended period of time')
-        logger.info('and cost additional fees')
+      if (errors) {
+        logger.info(`Errors have occurred while trying to release channels for ${market}`.red)
+        errors.forEach(e => logger.info(`- ${e}`))
+        logger.info('')
       }
 
+      if (errors && !force) {
+        logger.info('If the error above suggests an uncooperative closing of any channels on')
+        logger.info('your daemon, you can force the release of funds using the `--force` option.')
+        logger.info('Using `--force` has the potential to lock your funds for an extended')
+        logger.info('period of time and cost additional fees')
+      }
+
+      if (!errors) {
+        logger.info(`Successfully closed channels on ${market} market!`)
+      }
+    } catch (e) {
+      logger.error(`Failed to release payment channels for ${market}`.red)
       throw e
     }
-
-    logger.info(`Successfully closed channels on ${market} market!`)
   } catch (e) {
     logger.error(handleError(e))
   }
--- a/broker-cli/commands/wallet.js
+++ b/broker-cli/commands/wallet.js
@@ -358,7 +358,7 @@ async function release (args, opts, logger) {
 
       const { channels = [] } = await client.walletService.releaseChannels({ market, force })
 
-      if (!channels) {
+      if (!channels.length) {
         throw new Error('No information was retrieved from the daemon')
       }
 
--- a/broker-cli/commands/wallet.js
+++ b/broker-cli/commands/wallet.js
@@ -318,23 +318,38 @@ async function release (args, opts, logger) {
     if (!ACCEPTED_ANSWERS.includes(answer.toLowerCase())) return
 
     try {
-      const { errors } = await client.walletService.releaseChannels({ market, force })
+      let channelHasError = false
 
-      if (errors) {
+      const { channels = [] } = await client.walletService.releaseChannels({ market, force })
+
+      if (!channels) {
+        throw new Error('No information was retrieved from the daemon')
+      }
+
+      logger.info(`Released Market: ${market}`)
+
+      channels.forEach((channel) => {
+        const { symbol, hasError } = channel
+        let { status } = channel
+
+        if (hasError) {
+          channelHasError = true
+          logger.info(`${symbol}: ` + status.red)
+        } else {
+          logger.info(`${symbol}: ` + status.green)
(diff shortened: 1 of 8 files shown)


broker-cli/commands/wallet.js
```diff
@@ -318,20 +318,28 @@ async function release (args, opts, logger) {
...
+        errors.forEach(e => logger.info(e))
+        logger.info('')
+      }
 
-      if (!force) {
-        logger.info('You can force the release of funds using the `--force` option, however')
-        logger.info('this has the potential to lock your funds for an extended period of time')
-        logger.info('and cost additional fees')
+      if (errors && !force) {
+        logger.info('If the error above suggests an uncooperative closing of any channels on')
```
treygriffith: [nit] consider using a multi-line string here

treygriffith: is there no way for us to know if the error will suggest uncooperative closing?

broker-daemon/proto/broker.proto
```diff
@@ -925,6 +925,11 @@ message ReleaseChannelsRequest {
   bool force = 2;
 }
 
+message ReleaseChannelsResponse {
+  // Array of errors that may have occurred when releasing a specific market
+  repeated string errors = 1;
```
treygriffith: it seems like we ought to be returning more than just an array of strings. Perhaps a list of messages related to each engine? or to each channel? with success or failure.

broker-cli/commands/wallet.js
dannypaz: @treygriffith the formatting is easier w/ individual lines so I left it

ex:
```
if (error) {
  logger.info('without weird spacing')
}
```
vs
```
if (error) {
  logger.info(`
with weird spacing
  `)
}

Comment (dannypaz) [reply broker-cli/commands/wallet.js]: Per our talk, we will make `wallet release` an idempotent call (can be called multiple times). We will provide errors to the user and they can simply try the call again.

**Comment by dannypaz** [comment]
[releasing by market if one of the engines is down](https://trello.com/c/mUucp4Yv/921-releasing-by-market-if-one-of-the-engines-is-down)

File: broker-cli/commands/wallet.js [hunk 3]
```diff
@@ -352,20 +352,44 @@ async function release (args, opts, logger) {
...
-      await client.walletService.releaseChannels({ market, force })
-    } catch (e) {
-      logger.error(`Failed to release payment channels for ${market}`.red)
+      // We want to keep track if any returned channel had an error while attempting
+      // to be released so that we can display the proper error warnings below
+      let channelHasError = false
+
+      const { channels = [] } = await client.walletService.releaseChannels({ market, force })
+
+      if (!channels) {
```
treygriffith: this will never be true because you are defaulting `channels` to `[]` which is truthy.

broker-cli/commands/wallet.js
```diff
@@ -352,20 +352,44 @@ async function release (args, opts, logger) {
...
+        throw new Error('No information was retrieved from the daemon')
+      }
+
+      logger.info(`Released Market: ${market}`)
```
treygriffith: this log is a bit weird in the context of the logs that you do below, and that the user specifically requested this market be released.

broker-cli/commands/wallet.js
```diff
@@ -352,20 +352,44 @@ async function release (args, opts, logger) {
...
+          logger.info(`${symbol}: ` + status.green)
+        }
+      })
+
+      if (channelHasError) {
+        logger.info(`\nErrors have occurred while trying to release channels for ${market}:`.red)
       }
 
+      if (channelHasError && !force) {
+        logger.info('If the error above suggests an uncooperative closing of any channels on')
```
treygriffith: My comment from before still applies I think: is there any way to know if the error suggests an uncooperative close?

I'm imagining an output like this:

```
BTC: Released
LTC: Error: Peer is unavailable. You must wait for the peer to come online or perform an uncooperative close.

Errors have occurred while trying to release channels for BTC/LTC:
If the error above suggests an uncooperative closing of any channels on
your daemon, you can force the release of funds using the --force option.
```

Doesn't that seem a little weird?

broker-cli/proto/broker.proto
```diff
@@ -931,6 +931,23 @@ message ReleaseChannelsRequest {
   bool force = 2;
 }
 
+/**
+ * Container for channel information
+ */
+message ReleasedChannel {
```
treygriffith: calling this `ReleasedChannel` is a little weird since it could represent more than one channel.

broker-cli/commands/wallet.js
dannypaz: @treygriffith you think I should just remove it?

broker-cli/commands/wallet.js
dannypaz: @treygriffith the user has to make the decision to force close the channels, so RE: showing a message, there is no way around it. However, matching the error is an uphill battle, the only way we could potentially identify the error is with string matching

broker-cli/proto/broker.proto
dannypaz: @treygriffith What do you mean. This message type represents a single channel

broker-cli/commands/wallet.js
dannypaz: @treygriffith I went for it and made the change. it looks good

broker-cli/proto/broker.proto
```diff
@@ -931,6 +931,25 @@ message ReleaseChannelsRequest {
   bool force = 2;
 }
 
+/**
+ * Container for channel information for a specific currency on a given market
+ */
+message ReleasedCurrency {
+  // Currency of the information
+  Symbol symbol = 1;
```
treygriffith: I thought we were moving away from using the enum for `symbol` and using strings instead, right?

dannypaz: @treygriffith removed it with the last commit

PR #302, opened 2018-12-04
check funds in placeorder and fillorder when the counterAmount will a…

martineehrlich:
…lways exist

## Description
A few sentences describing the overall goals of the pull request's commits.

## Related PRs

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Link to Trello

### DIFF
--- a/broker-daemon/block-order-worker/index.js
+++ b/broker-daemon/block-order-worker/index.js
@@ -99,28 +99,7 @@ class BlockOrderWorker extends EventEmitter {
       throw new Error(`No engine available for ${orderbook.counterSymbol}.`)
     }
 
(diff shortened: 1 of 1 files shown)


broker-daemon/block-order-worker/index.js
```diff
@@ -466,6 +472,8 @@ class BlockOrderWorker extends EventEmitter {
       throw new Error(`No engine available for ${counterSymbol}`)
     }
 
+    await this.checkFundsAreSufficient(blockOrder)
```
treygriffith: you'll need to do this for each fill, since even here the block order doesn't have a `counterAmount`

PR #303, opened 2018-12-05
Bugfix/counter amount should exist

martineehrlich:
## Description
With the new check for sufficient funds, we have a case where the counterAmount to check with does not exist (if the order is a market order). 
This PR defines a function that finds the average weighted price so that we can approximate a counter amount when checking if the engine has sufficient funds.

## Todos
- [x] Tests
- [x] Documentation
- [x] Link to Trello

### DIFF
--- a/broker-daemon/block-order-worker/index.js
+++ b/broker-daemon/block-order-worker/index.js
@@ -99,27 +99,49 @@ class BlockOrderWorker extends EventEmitter {
       throw new Error(`No engine available for ${orderbook.counterSymbol}.`)
     }
 
+    const blockOrder = new BlockOrder({ id, marketName, side, amount, price, timeInForce })
+    await this.checkFundsAreSufficient(blockOrder, orderbook)
+
+    await promisify(this.store.put)(blockOrder.key, blockOrder.value)
+
+    this.logger.info(`Created and stored block order`, { blockOrderId: blockOrder.id })
+
+    // Start working the block order asynchronously to prevent blocking the creation
+    // of 'other' block orders
+    this.workBlockOrder(blockOrder, Big(blockOrder.baseAmount)).catch(err => {
+      this.failBlockOrder(blockOrder.id, err)
+    })
+
+    return id
+  }
+
+  /**
+   * Checks that there are valid inbound and outbound funds to place/fill the order
+   *
+   * @param {BlockOrder}
+   * @param {Orderbook}
+   * @return {Void}
+   * @throws {Error} If there are insufficient outbound or inbound funds
+   */
+  async checkFundsAreSufficient (blockOrder, orderbook) {
+    const { marketName, side, outboundSymbol, inboundSymbol } = blockOrder
     const { activeOutboundAmount, activeInboundAmount } = await this.calculateActiveFunds(marketName, side)
 
-    const blockOrder = new BlockOrder({ id, marketName, side, amount, price, timeInForce })
-    const outboundEngine = this.engines.get(blockOrder.outboundSymbol)
-    const inboundEngine = this.engines.get(blockOrder.inboundSymbol)
+    const outboundEngine = this.engines.get(outboundSymbol)
+    const inboundEngine = this.engines.get(inboundSymbol)
     const [{address: outboundAddress}, {address: inboundAddress}] = await Promise.all([
-      this.relayer.paymentChannelNetworkService.getAddress({symbol: blockOrder.outboundSymbol}),
-      this.relayer.paymentChannelNetworkService.getAddress({symbol: blockOrder.inboundSymbol})
+      this.relayer.paymentChannelNetworkService.getAddress({symbol: outboundSymbol}),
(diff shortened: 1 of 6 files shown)


broker-daemon/block-order-worker/index.js
```diff
@@ -109,14 +109,37 @@ class BlockOrderWorker extends EventEmitter {
...
+    let counterAmount
+    let outboundAmount
+    let inboundAmount
+    if (blockOrder.isMarketOrder) {
+      const { orders, depth } = await orderbook.getBestOrders({ side: blockOrder.inverseSide, depth: Big(blockOrder.baseAmount).toString() })
+      if (Big(depth).lt(blockOrder.baseAmount)) {
+        this.logger.error(`Insufficient depth`)
+        throw new Error(`Insufficient depth`)
+      }
+      counterAmount = await orderbook.getAveragePrice(orders, Big(depth))
```
treygriffith: don't convert to `Big` here, `getAveragePrice` should be responsible for that

broker-daemon/block-order-worker/index.js
```diff
@@ -109,14 +109,37 @@ class BlockOrderWorker extends EventEmitter {
       this.relayer.paymentChannelNetworkService.getAddress({symbol: blockOrder.inboundSymbol})
     ])
 
-    const outboundBalanceIsSufficient = await outboundEngine.isBalanceSufficient(outboundAddress, Big(blockOrder.outboundAmount).plus(activeOutboundAmount))
+    let counterAmount
+    let outboundAmount
+    let inboundAmount
+    if (blockOrder.isMarketOrder) {
+      const { orders, depth } = await orderbook.getBestOrders({ side: blockOrder.inverseSide, depth: Big(blockOrder.baseAmount).toString() })
```
treygriffith: why not just use `blockOrder.baseAmount` for depth (instead of doing `Big(...).toString()`?

broker-daemon/block-order-worker/index.js
```diff
@@ -109,14 +109,37 @@ class BlockOrderWorker extends EventEmitter {
       this.relayer.paymentChannelNetworkService.getAddress({symbol: blockOrder.inboundSymbol})
     ])
 
-    const outboundBalanceIsSufficient = await outboundEngine.isBalanceSufficient(outboundAddress, Big(blockOrder.outboundAmount).plus(activeOutboundAmount))
+    let counterAmount
+    let outboundAmount
+    let inboundAmount
+    if (blockOrder.isMarketOrder) {
```
treygriffith: add a comment about why you are doing this

broker-daemon/block-order-worker/index.js
```diff
@@ -109,14 +109,37 @@ class BlockOrderWorker extends EventEmitter {
...
+      } else {
+        outboundAmount = blockOrder.outboundAmount
+        inboundAmount = counterAmount
+      }
+    } else {
+      outboundAmount = blockOrder.outboundAmount
+      inboundAmount = blockOrder.inboundAmount
+    }
+
+    const outboundBalanceIsSufficient = await outboundEngine.isBalanceSufficient(outboundAddress, Big(outboundAmount).plus(activeOutboundAmount))
```
treygriffith: I would consider moving this sufficiency check into its own method, as `createBlockOrder` is getting pretty unwieldy

broker-daemon/orderbook/index.js
```diff
@@ -217,6 +217,28 @@ class Orderbook {
...
   }
 
+  async getAveragePrice (orders, targetDepth) {
+    targetDepth = Big(targetDepth)
+    let currentDepth = Big(0)
+    let weightedPrice = Big(0)
+    orders.forEach((order) => {
+      const depthRemaining = targetDepth.minus(currentDepth)
+
+      // if we have already reached our target depth, create no further fills
```
treygriffith: update comment

broker-daemon/orderbook/index.js
```diff
@@ -217,6 +217,28 @@ class Orderbook {
     })
   }
 
+  async getAveragePrice (orders, targetDepth) {
```
treygriffith: `getAveragePrice` should probably just retrieve the orders directly unless there is a good reason for it not to

broker-daemon/orderbook/index.js
```diff
@@ -217,6 +217,28 @@ class Orderbook {
...
+        return
+      }
+
+      // Take the smaller of the remaining desired depth or the base amount of the order
+      const fillAmount = depthRemaining.gt(order.baseAmount) ? order.baseAmount : depthRemaining.toString()
+      // track our current depth so we know what to fill on the next order
+      currentDepth = currentDepth.plus(fillAmount)
+      weightedPrice = weightedPrice.plus(order.price.times(fillAmount))
+    })
+
```
treygriffith: should handle if your `currentDepth` does not reach your `targetDepth`

PR #305, opened 2018-12-08
Fix - Multi-engine support for network status

dannypaz:
## Description
If a single daemon is down, the network-status call will fail to display information for all currencies. We should make the `wallet network-status` call resilient to engine failures to show information for daemons that are running.

screen shot 2018-12-07 at 6 08 16 pm

## Related PRs

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Link to Trello

### DIFF
--- a/broker-cli/commands/wallet.js
+++ b/broker-cli/commands/wallet.js
@@ -250,8 +250,11 @@ const NETWORK_STATUSES = Object.freeze({
  * @param {String} status
  * @return {String} balance - formatted with size and color
  */
-function formatBalance (balance, status) {
-  if (!balance) {
+function formatBalance (balance, status, errorPresent) {
+  // If there were errors when receiving balances, the balance will come back
+  // as either a blank string or null. We set the balance to `Not Available` in this
+  // case, but should make sure we dont treat `0` as the same type.
+  if (balance === '' || balance === null) {
     return 'Not Available'.yellow
   }
 
@@ -272,6 +275,15 @@ function formatBalance (balance, status) {
 
   return fixedBalance
 }
+
+/**
+ * @constant
+ * @type {Object<key, String>}
+ * @default
+ */
+const CAPACITY_STATUSES = Object.freeze({
+  OK: 'OK'
+})
 
 /**
  * network-status
@@ -302,7 +314,7 @@ async function networkStatus (args, opts, logger) {
     })
 
     // If any balances in the following table are empty (which occurs if the engine is unavailable)
-    // then the text will show up as `N/A`.
+    // then the text will show up as `Not Available`.
     //
     // The user will then be prompted with a warning letting them know that we failed
     // to receive a balance for a particular currency
@@ -326,11 +338,11 @@ async function networkStatus (args, opts, logger) {
     logger.info(statusTable.toString())
     logger.info('')
 
-    if (baseSymbolCapacities.status !== 'OK') {
+    if (baseSymbolCapacities.status !== CAPACITY_STATUSES.OK) {
       logger.error(`${baseSymbol}: Received errors when requesting network status: ${baseSymbolCapacities.error}`.red)
     }
 
-    if (counterSymbolCapacities.status !== 'OK') {
+    if (counterSymbolCapacities.status !== CAPACITY_STATUSES.OK) {
       logger.error(`${counterSymbol}: Received errors when requesting network status: ${counterSymbolCapacities.error}`.red)
     }
   } catch (e) {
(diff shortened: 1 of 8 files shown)


broker-cli/commands/wallet.js
```diff
@@ -241,6 +241,38 @@ const NETWORK_STATUSES = Object.freeze({
   INACTIVE: 'INACTIVE'
 })
 
+/**
```
dannypaz: Moved this function above the point where it is called and added in a check for `if (!balance)`

broker-cli/commands/wallet.js
```diff
@@ -261,62 +293,49 @@ async function networkStatus (args, opts, logger) {
     const client = new BrokerDaemonClient(rpcAddress)
     const { baseSymbolCapacities, counterSymbolCapacities } = await client.walletService.getTradingCapacities({market})
 
+    const baseSymbol = baseSymbolCapacities.symbol.toUpperCase()
+    const counterSymbol = counterSymbolCapacities.symbol.toUpperCase()
+
     const statusTable = new Table({
-      head: ['', `${baseSymbolCapacities.symbol.toUpperCase()} Capacity`, `${counterSymbolCapacities.symbol.toUpperCase()} Capacity`],
-      colWidths: [14, 24, 24],
+      head: ['', `${baseSymbol} Capacity`, `${counterSymbol} Capacity`],
```
dannypaz: Removed column widths because Table will auto scale to the values in the columns

broker-cli/proto/broker.proto
```diff
@@ -889,6 +889,8 @@ message Capacity {
   string outstanding_send_capacity = 10;
   // Total outbound outstanding amount. This is currently in base units (e.g. Satoshis/Litoshis)
   string outstanding_receive_capacity = 11;
+  // true/false for any errors that may have occurred when requesting capacity for a particular currency
+  bool error = 20;
```
dannypaz: We dont require an error message (we simply want to know if something happened) but not sure if this is the correct way to go about it

dannypaz: I changed my mind on this and added a `status` and `error` prop used for messages. This fits inline with other endpoints

broker-daemon/broker-rpc/wallet-service/release-channels.js
```diff
@@ -9,6 +9,7 @@ const RELEASE_STATE = Object.freeze({
   RELEASED: 'RELEASED',
   FAILED: 'FAILED'
 })
+
```
dannypaz: Added a space because I copy pasta-ed from this file

dannypaz:
[network status if one engine is down](https://trello.com/c/R98HLUru/920-network-status-if-one-engine-is-down)

broker-cli/commands/wallet.js
```diff
@@ -241,6 +241,38 @@ const NETWORK_STATUSES = Object.freeze({
...
+ * Formats the capacities to fixed numbers and colors them if they are greater
+ * than 0. Will return Not Available if balance is empty
+ *
+ * @function
+ * @param {String} balance
+ * @param {String} status
+ * @return {String} balance - formatted with size and color
+ */
+function formatBalance (balance, status) {
+  if (!balance) {
```
treygriffith: I'm not sure we want to use the balance not existing as evidence that the engine is unavailable - any way to make this more explicit.

If we are to keep it this way, I'd prefer to do something like `balance === undefined`

broker-cli/commands/wallet.js
```diff
@@ -261,62 +293,49 @@ async function networkStatus (args, opts, logger) {
...
-        return fixedBalance.green
-      case NETWORK_STATUSES.OUTSTANDING:
-        return fixedBalance.yellow
-      case NETWORK_STATUSES.PENDING:
-        return fixedBalance.yellow
-      case NETWORK_STATUSES.INACTIVE:
-        return fixedBalance.red
-      default:
-        return fixedBalance
+    if (baseSymbolCapacities.status !== 'OK') {
```
treygriffith: string for `OK` is a little strange

dannypaz: @treygriffith want me to move to an enum?

broker-cli/commands/wallet.js
dannypaz: @treygriffith it will come across as an empty string through gRPC. Unfortunately there is not a pretty way to do this. The code that uses `formatBalance` is massive.

broker-cli/commands/wallet.js
dannypaz: Ive moved the string into an enum

PR #308, opened 2018-12-10
Feature - Fee Estimation to wallet commit

dannypaz:
## Description
We currently have commit limits to BTC and LTC when a "user" uses the broker. When a user tries to commit their currency under these limits, there are times when the fees are too high and do not cover the opening transaction

This PR adds a very archaic "fee estimation" that provides a buffer for the user so that we can commit all of their available currency with respect to fees. 

We've set conservative, hardcoded limits for each currency. In the future we will expose fee estimation through the engine to provide realtime limits. 

## Related PRs

## Todos
- [x] Tests
- [x] Documentation
- [x] Link to Trello


dannypaz:
[Fee estimation for channel opening](https://trello.com/c/IOBDV6Kk/723-fee-estimation-for-channel-opening)

broker-daemon/broker-rpc/wallet-service/commit.js
```diff
@@ -12,25 +12,43 @@ const { currencies: currencyConfig } = require('../../config')
...
   }
 
+  // A very archaic fee estimation amount. This number was chosen based on fees and
+  // pricing of bitcoin at the current time (Dec 7 pricing $3.5k).
+  //
+  // As a conservative number, we would typically expect fees to be between $0.10 and $0.40
+  // per on-chain transaction, so if we buffer our commitment amount by 10000sat ($0.40 usd)
+  // then we should be relatively safe.
+  //
+  // The equivalent default for litecoin would be around 1000 litoshis
```
treygriffith: I think these litecoin fees are low

broker-daemon/broker-rpc/wallet-service/commit.js
```diff
@@ -112,8 +134,10 @@ async function commit ({ params, relayer, logger, engines, orderbooks }, { Empty
     logger.debug('Requesting inbound channel from relayer', { address: paymentChannelNetworkAddress, balance: convertBalance, symbol: inverseSymbol })
     await relayer.paymentChannelNetworkService.createChannel({address: paymentChannelNetworkAddress, balance: convertedBalance, symbol: inverseSymbol})
   } catch (e) {
-    // TODO: Close channel that was open if relayer call has failed
-    throw (e)
+    // If the relayer call fails, the user can simply try and open channels again
+    // on the same market. We handle this undesired state above where we check the
+    // current status of channels and repair them if necessary.
+    throw new PublicError(`Funding error: Relayer is unavailable, Please try again`)
```
treygriffith: not sure that `Relayer is unavailable` is the message I would send here. At the very least we should log the actual error, and given an error like `Relayer error`.

broker-daemon/broker-rpc/wallet-service/commit.js
dannypaz: @treygriffith definitely low. I've updated the amounts to `$0.50 USD`

broker-daemon/broker-rpc/wallet-service/commit.js
```diff
@@ -97,10 +117,14 @@ async function commit ({ params, relayer, logger, engines, orderbooks }, { Empty
     throw new PublicError(errorMessage)
   }
 
-  logger.debug('Creating outbound channel', { address, balance })
+  logger.debug('Creating outbound channel', { address, balance, feeEstimate })
+
+  // We remove fees from the balance to make sure that the user has enough funds
+  // for the open/close channel transactions
+  const balanceWithFeeEstimate = Big(balance).minus(feeEstimate).toString()
```
treygriffith: I think this should probably happen in the engine itself

dannypaz:
@treygriffith going to move this PR to the engine

PR #311, opened 2018-12-10
Feature: use a custom logger instead of message ID in grpc methods

treygriffith:
## Description
We need to filter out some data from certain logs to prevent sensitive information from being accidentally persisted.

Rather than further complicate grpc methods, this change uses a custom logger object to do filtering our behalf.

This is a proof-of-concept of the approach.

## Related PRs
https://github.com/sparkswap/grpc-methods/pull/13

## Todos
- [x] AdminService
- [ ] InfoService
- [ ] OrderService
- [ ] OrderbookService
- [ ] WalletService
- [ ] Filter cipher payload


treygriffith:
closing in favor of #312

PR #312, opened 2018-12-11
Filter sensitive logs

treygriffith:
## Description
This PR filters out sensitive information from the broker logs so as to avoid accidentally saving sensitive data in an insecure format.

## Related PRs
#311 was an another attempt at this.

### DIFF
--- a/broker-daemon/utils/logger.js
+++ b/broker-daemon/utils/logger.js
@@ -14,6 +14,7 @@
 const winston = require('winston')
 
 const sensitiveList = [
+  'username',
   'pass',
   'password',
   'passphrase',
--- a/broker-daemon/utils/logger.js
+++ b/broker-daemon/utils/logger.js
@@ -13,11 +13,28 @@
 
 const winston = require('winston')
 
+const sensitiveList = [
+  'pass',
+  'password',
+  'passphrase',
+  'recoverySeed'
+]
+const filterSensitive = winston.format((info, opts) => {
+  return sensitiveList.reduce(([info, key]) => {
+    const updatedInfo = Object.assign({}, info)
+    if (updatedInfo[key] != null) {
+      updatedInfo[key] = '***FILTERED***'
+    }
+    return updatedInfo
+  }, info)
+});
+
 const logger = winston.createLogger({
   level: (process.env.NODE_ENV === 'production') ? 'info' : 'debug',
   format: winston.format.combine(
+    filterSensitive(),
     winston.format.timestamp(),
-    winston.format.json()
+    winston.format.json(),
   ),
   json: true,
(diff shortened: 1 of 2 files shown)


broker-daemon/utils/logger.js
```diff
@@ -13,9 +13,27 @@
 
 const winston = require('winston')
 
+const sensitiveList = [
```
dannypaz: [nit] make uppercase `SENSITIVE_WORD_LIST` and add `Object.freeze([...`

broker-daemon/utils/logger.js
```diff
@@ -13,9 +13,27 @@
...
+  'username',
+  'pass',
+  'password',
+  'passphrase',
+  'recoverySeed'
+]
+const filterSensitive = winston.format((info, opts) => {
+  return sensitiveList.reduce((info, key) => {
```
dannypaz: using `info` in 3 places is a little confusing.

broker-daemon/utils/logger.js
```diff
@@ -13,9 +13,27 @@
...
+    const updatedInfo = Object.assign({}, info)
+    if (updatedInfo[key] != null) {
+      updatedInfo[key] = '***FILTERED***'
+    }
+    return updatedInfo
+  }, info)
```
dannypaz: I think this needs to be `opts` instead?

broker-daemon/utils/logger.js
```diff
@@ -13,9 +13,27 @@
...
+  'pass',
+  'password',
+  'passphrase',
+  'recoverySeed'
+]
+const filterSensitive = winston.format((info, opts) => {
+  return sensitiveList.reduce((info, key) => {
+    const updatedInfo = Object.assign({}, info)
+    if (updatedInfo[key] != null) {
+      updatedInfo[key] = '***FILTERED***'
```
martineehrlich: This string should be a constant

broker-daemon/utils/logger.js
```diff
@@ -13,9 +13,41 @@
...
+ * @constant
+ * @type {String}
+ */
+const SENSITIVE_REPLACEMENT = '***FILTERED***'
+
+/**
+ * Formatter to filter sensitive data from logs.
+ */
+const filterSensitive = winston.format((info) => {
+  return SENSITIVE_PROP_LIST.reduce((info, key) => {
```
martineehrlich: Okay this might be annoying, but what initialize and empty object and if you iterate through the info object instead of the SENSITIVE_PROP_LIST.

That way there is no mutation and you perform the same behavior on all the key values, and just filter when the key is in the SENSITIVE_PROP_LIST?

PR #313, opened 2018-12-11
throw error if get balances returns an error

martineehrlich:
## Description
There is a bug in commit where we call Big on an empty string if the engine is down because the getBalances call returns empty strings and an error. Before calling Big on the uncommitted balance, we should see if there is an error in the balances object, and throw an error if there is.

## Related PRs

## Todos
- [x] Tests
- [x] Documentation
- [x] Link to Trello

### DIFF
--- a/broker-cli/commands/wallet.js
+++ b/broker-cli/commands/wallet.js
@@ -154,7 +154,11 @@ async function commit (args, opts, logger) {
 
     const { balances } = await client.walletService.getBalances({})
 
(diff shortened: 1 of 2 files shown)


martineehrlich:
[bigjs error if engine is down while trying to commit balance](https://trello.com/c/Q0x9HBof/949-bigjs-error-if-engine-is-down-while-trying-to-commit-balance)

PR #315, opened 2018-12-13
Feature/recreate state machines on startup

martineehrlich:
## Description
When the broker goes down, we can end up with orders/fills in unresolved states. This PR takes care of this by rehydrating those orders/fills on blockOrderWorker initialization and triggering them to run from their current state to completion.

## Todos
- [x] Tests
- [x] Documentation
- [x] Link to Trello

### DIFF
--- a/broker-daemon/block-order-worker/index.js
+++ b/broker-daemon/block-order-worker/index.js
@@ -74,28 +74,8 @@ class BlockOrderWorker extends EventEmitter {
 
   async settleIndeterminateOrdersFills () {
     const blockOrders = await getRecords(this.store, BlockOrder.fromStorage.bind(BlockOrder))
-
     for (let blockOrder of blockOrders) {
-      const orderStateMachines = await getRecords(
-        this.ordersStore,
-        (key, value) => {
-          return OrderStateMachine.fromStore(
-            {
-              store: this.ordersStore,
-              logger: this.logger,
-              relayer: this.relayer,
-              engines: this.engines
-            },
-            {
-              key,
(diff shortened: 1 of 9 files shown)


broker-daemon/block-order-worker/index.js
```diff
@@ -69,6 +69,63 @@ class BlockOrderWorker extends EventEmitter {
...
+                value
+              }
+            )
+          }
+        },
+        // limit the orders we retrieve to those that belong to this blockOrder, i.e. those that are in
+        // its prefix range.
+        Fill.rangeForBlockOrder(blockOrder.id)
+      )
+      for (let fsm in fillStateMachines) {
```
treygriffith: this should be `for...of`

treygriffith:
To recap our offline conversation:
- `goto` should not have any side effects, and should be safely applied to every state machine, regardless of what state we are moving to
- The retry should be triggered by the block order worker directly
- we should only apply listeners to those state machines that we are retrying

broker-daemon/state-machines/fill-state-machine.js
```diff
@@ -116,7 +118,12 @@ const FillStateMachine = StateMachine.factory({
      * execute transition: execute the swap itself
      * @type {Object}
      */
-    { name: 'execute', from: 'filled', to: 'executed' }
+    { name: 'execute', from: 'filled', to: 'executed' },
+    /**
+     * execute transition: execute the swap itself
```
treygriffith: wrong jsdoc

PR #327, opened 2018-12-15
Updating post install script to generate auto completion methods

mdasif:
## Description
Updating post install script to generate auto completion methods for shell bash and zsh.
https://trello.com/c/ONT5ioQ5/207-add-autocomplete-to-kcli

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Link to Trello

### DIFF
--- a/broker-cli/scripts/postinstall.sh
+++ b/broker-cli/scripts/postinstall.sh
@@ -1,5 +1,10 @@
 #!/bin/sh
 
+# Setup auto-completion for bash
+echo "source <(sparkswap completion bash)" >> ~/.bashrc && source ~/.bashrc
+# Setup auto-completion for zsh
+echo "source <(sparkswap completion zsh)" >> ~/.zshrc && source ~/.zshrc
+
 echo "Successfully installed Ϟ SparkSwap Broker CLI (sparkswap)!"
 echo ""
 echo "To set up your custom configuration for the SparkSwap CLI, copy 'sample-.sparkswap.js' to your home directory as '.sparkswap.js'."


broker-cli/scripts/postinstall.sh
```diff
@@ -1,5 +1,10 @@
 #!/bin/sh
 
+# Setup auto-completion for bash
+echo "source <(sparkswap completion bash)" >> ~/.bashrc && source ~/.bashrc
```
treygriffith: do we need to check that `bashrc` checks first? Or is there another way to determine the default shell?

I'd hate to populate a `.bashrc` if the user doesn't use `bash`.

dannypaz:
[Remove `--no-seedbackup` configuration](https://trello.com/c/1VoOmUsh/853-remove-no-seedbackup-configuration)

broker-cli/scripts/postinstall.sh
```diff
@@ -1,5 +1,14 @@
 #!/bin/sh
 
+# Setup auto-completion for bash
+if [-f ~/.bashrc]; then
+    echo "source <(sparkswap completion bash)" >> ~/.bashrc && source ~/.bashrc
```
dannypaz: This line will add that text each time `npm i` is ran, which is not necessarily what we want. Instead of adding the line automatically in the script, we should prompt the use to add it manually IF they want auto completion.

dannypaz:
@mdasif in order for auto-completion to work, we need to add `.complete` information to each cli call. An example for auto completion in caporal [can be found here](https://www.npmjs.com/package/caporal#auto-completion-setup-example)

broker-cli/scripts/postinstall.sh
```diff
@@ -6,4 +6,13 @@ echo "To set up your custom configuration for the SparkSwap CLI, copy 'sample-.s
 echo ""
 echo "To do this now, run the below (in bash):"
 echo "cp -n \"$(dirname $(which sparkswap))/../lib/node_modules/broker-cli/sample-.sparkswap.js\" ~/.sparkswap.js"
+echo ""
+echo "You can also enable autocompletion of Sparkswap CLI."
+echo ""
+echo "If your favorite shell is bash, run the below command:"
+echo "echo \"source <(sparkswap completion bash)\" >> ~/.bashrc && source ~/.bashrc"
+echo ""1:
```
dannypaz: Did you mean to have `1:` here?

dannypaz:
[Add autocomplete to sparkswap](https://trello.com/c/ONT5ioQ5/207-add-autocomplete-to-sparkswap)

PR #328, opened 2018-12-15
Fix/healthcheck

dannypaz:
## Description
This PR adds status codes to `getEngineStatus` and refactors the call for changes made in the lnd-engine at https://github.com/sparkswap/lnd-engine/pull/137

## Related PRs

## Todos
- [x] Tests
- [x] Documentation
- [x] Link to Trello


broker-cli/commands/health-check.js
```diff
@@ -10,7 +10,8 @@ const { validations, handleError } = require('../utils')
  */
 const STATUS_CODES = Object.freeze({
   OK: 'OK',
-  UNKNOWN: 'UNKNOWN'
+  UNKNOWN: 'UNKNOWN',
+  VALIDATED: 'VALIDATED'
```
treygriffith: seems weird to define this is a string when it should be available on the engine itself, right?

dannypaz: @treygriffith this is definitely weird, but what this does is change the `VALIDATED` status to `OK`, since it was really weird to have a term that the user might not understand?

dannypaz: Its essentially a UI change https://github.com/sparkswap/broker/pull/328/files#diff-54e861896da44d61a844920c19b0bc2eR46

dannypaz: @treygriffith maybe the way to go is to change the `VALIDATED` status code to `OK` on the lnd-engine?

broker-cli/commands/health-check.js
```diff
@@ -1,3 +1,4 @@
+const LndEngine = require('lnd-engine')
```
treygriffith: can you add a todo in here to make this reference a more generic implementation?

PR #329, opened 2018-12-17
Fixing SparkSwap CLI commands order

mdasif:
* Adding descriptions.
* Fix ordering
* Remove dot from all help lines.

## Todos
- [x] Tests
- [ ] Documentation
- [x] Link to Trello https://trello.com/c/CQAGXY38/250-ordering-of-commands-on-sparkswap-cli


broker-cli/commands/buy.js
```diff
@@ -47,11 +47,11 @@ async function buy (args, opts, logger) {
 
 module.exports = (program) => {
   program
-    .command('buy', 'Submit an order to buy.')
-    .argument('<amount>', 'Amount of base currency to buy.', validations.isDecimal)
-    .argument('[price]', 'Worst price that this order should be executed at. (If omitted, the market price will be used)', validations.isDecimal)
+    .command('buy', 'Submit an order to buy')
+    .argument('<amount>', 'Amount of base currency to buy', validations.isDecimal)
+    .argument('[price]', 'Worst price that this order should be executed at (If omitted, the market price will be used)', validations.isDecimal)
```
treygriffith: Let's not end this sentence with a preposition (`at`)

broker-cli/commands/identity.js
```diff
@@ -31,6 +31,6 @@ async function getIdentity (args, opts, logger) {
 module.exports = (program) => {
   program
     .command('id', 'Gets the Public Key of the Broker Daemon')
-    .option('--rpc-address [rpc-address]', 'Location of the RPC server to use.', validations.isHost)
+    .option('--rpc-address [rpc-address]', 'Location of the RPC server to use', validations.isHost)
```
treygriffith: is it worthwhile to consolidate this option somewhere so that we have consistent language across the commands?

broker-cli/commands/info/index.js
```diff
@@ -50,12 +50,12 @@ module.exports = (program) => {
       }
     })
     .command(`info ${SUPPORTED_COMMANDS.SUPPORTED_MARKETS}`, 'Get the markets currently supported')
-    .option('--rpc-address [rpc-address]', 'Location of the RPC server to use.', validations.isHost)
-    .command(`info ${SUPPORTED_COMMANDS.MARKET_STATS}`, 'Get the markets currently supported')
+    .option('--rpc-address [rpc-address]', 'Location of the RPC server to use', validations.isHost)
+    .command(`info ${SUPPORTED_COMMANDS.MARKET_STATS}`, 'Get statistics (price ticker information) for a particular market/symbol for a period of the last 24 hours')
```
treygriffith: I think we should use `market` (instead of symbol) here

broker-cli/commands/info/index.js
```diff
@@ -50,12 +50,12 @@ module.exports = (program) => {
...
     .option('--market [marketName]', 'Relevant market name')
-    .command(`info ${SUPPORTED_COMMANDS.TRADES}`, 'Get the markets currently supported')
-    .argument('<since>', 'Datetime for lowerbound of range.', validations.isDate)
+    .command(`info ${SUPPORTED_COMMANDS.TRADES}`, 'Get detailed information about trades from a given time range')
+    .argument('<since>', 'Datetime for lowerbound of range', validations.isDate)
```
treygriffith: this description isn't too helpful. I would consider re-wording this to be something more readable.

broker-cli/commands/identity.js
mdasif: I think that's a good idea. as a side note,why do we have an option to pass `rpc-address`. isn't `rpc-address` and `certs` a pair which will be setup at the time of installation of client & either both should be configurable or none.

dannypaz: @mdasif the setup is a little rough. In the past (before the TLS requirement for cli), you could simply specific another broker address with `rpc-address`. It is still possible to have multiple certs and use this command, however we should probably add a `--rpc-cert` tag that allows the user to specify a cert.

treygriffith:
@mdasif can you include a picture of the new CLI help screen?

PR #330, opened 2018-12-17
Feature/enable auth

martineehrlich:
## Description
In order to enable SSL/TLS we needed to turn the disableAuth flag to false and then also update where the certs live and how we install the config file to the right place.
1) The broker-daemon build script now grabs the certs from the secure folder and puts them into the sparkswap file in the home directory
2) The broker-cli has an install-config script which puts the sparkswap config in the sparkswap directory if one does not already exist there.
3) The broker-cli postinstall script points the user to use the install-config script

Install and postinstall scripts

## Related PRs
https://github.com/sparkswap/docs/pull/21

## Todos
- [x] Tests
- [x] Documentation
- [x] Link to Trello


broker-cli/broker-daemon-client/index.js
```diff
@@ -73,8 +74,12 @@ class BrokerDaemonClient {
       if (!this.username) throw new Error('No username is specified for authentication')
       if (!this.password) throw new Error('No password is specified for authentication')
 
-      // Go back to the ./broker-cli/certs directory from the current directory
-      this.cert = readFileSync(path.join(PROJECT_ROOT, this.certPath))
+      let certPathParts = this.certPath.split(path.sep)
+      if (certPathParts[0] === '~') {
+        certPathParts[0] = os.homedir()
+      }
+      const certPath = path.join(...certPathParts)
```
dannypaz: [nit] I dont think we need to spread here, since path.join should take the array

broker-cli/broker-daemon-client/index.js
```diff
@@ -73,8 +74,12 @@ class BrokerDaemonClient {
       if (!this.username) throw new Error('No username is specified for authentication')
       if (!this.password) throw new Error('No password is specified for authentication')
 
-      // Go back to the ./broker-cli/certs directory from the current directory
-      this.cert = readFileSync(path.join(PROJECT_ROOT, this.certPath))
+      let certPathParts = this.certPath.split(path.sep)
```
dannypaz: would add a comment JIC to let the dev know that we are trying to resolve the tilde

broker-cli/scripts/install-config.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/sh
```
dannypaz: [nit] unless there is a reason to use sh, most of our scripts use `#!/usr/bin/env bash`

martineehrlich: Ah I was just being consistent with the postinstall script in broker-cli, but I'll change.

broker-cli/broker-daemon-client/index.js
dannypaz: this is so wrong ^^ it takes segments, not an array. sorry @martineehrlich

scripts/build.sh
```diff
@@ -56,3 +56,15 @@ fi
 
 echo "Building broker docker images"
 EXTERNAL_ADDRESS=$EXTERNAL_ADDRESS npm run build-images
+
+echo "Copying certs"
+DIRECTORY=~/.sparkswap
```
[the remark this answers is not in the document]
martineehrlich: I don't understand this comment

dannypaz:
@martineehrlich can you add a screenshot of the postinstall script to the description of the PR? Also TY for the changes. LGTM, but waiting for @treygriffith to rereview

martineehrlich:
[Make DISABLE_AUTH false by default](https://trello.com/c/263hrJFt/840-make-disableauth-false-by-default)

broker-cli/broker-daemon-client/index.spec.js
```diff
@@ -98,25 +98,48 @@ describe('BrokerDaemonClient', () => {
...
+      BrokerDaemonClient.__set__('path', { join: joinStub, sep: '/' })
+      BrokerDaemonClient.__set__('os', osStub)
     })
 
     it('reads a cert file', () => {
       broker = new BrokerDaemonClient()
       expect(readFileSyncStub).to.have.been.calledWith(certPath)
     })
 
+    it('expands the filepath if it is pointing to root', () => {
```
treygriffith: `home`, not `root`

broker-cli/broker-daemon-client/index.spec.js
```diff
@@ -98,25 +98,48 @@ describe('BrokerDaemonClient', () => {
...
       loadConfigStub.returns({
         rpcAddress: address,
-        rpcCert: certPath,
+        rpcCertPath: certPath,
         disableAuth: false,
         rpcUser,
         rpcPass
       })
+
+      homedir = '/home/'
```
treygriffith: `os.homedir()` would return `/home` not `/home/` (no trailing slash)

PR #331, opened 2018-12-18
add logging to get-preimage

treygriffith:
## Description
This change adds some additional debug logging to the Interchain Router to assist with debugging since it is frequently the most difficult component to trace.

## Todos
- [ ] ~~Tests~~ (Logging change only)
- [ ] ~~Documentation~~

### DIFF
--- a/broker-daemon/interchain-router/get-preimage.js
+++ b/broker-daemon/interchain-router/get-preimage.js
@@ -83,6 +83,8 @@ async function getPreimage ({ params, send, onCancel, onError, ordersByHash, eng
   const { inboundSymbol, inboundFillAmount, outboundSymbol, outboundFillAmount, takerAddress } = order
   const [ expectedSymbol, actualSymbol, expectedAmount, actualAmount ] = [ inboundSymbol, symbol, inboundFillAmount, amount ]
 
+  logger.debug(`Found routing entry for swap ${swapHash}`)
+
   const outboundEngine = engines.get(outboundSymbol)
   if (!outboundEngine) {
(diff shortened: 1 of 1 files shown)


broker-daemon/interchain-router/get-preimage.js
```diff
@@ -92,7 +94,9 @@ async function getPreimage ({ params, send, onCancel, onError, ordersByHash, eng
 
   // We don't want to reject an incoming HTLC if we know that there is an active outgoing one. If the outgoing HTLC is in flight or completed,
   // we should attempt to retrieve the associated preimage.
+  logger.debug(`Checking outbound HTLC status for swap ${swapHash}`)
```
dannypaz: would it be worth it to include inbound/outbound symbols?

broker-daemon/interchain-router/get-preimage.js
```diff
@@ -83,6 +83,8 @@ async function getPreimage ({ params, send, onCancel, onError, ordersByHash, eng
   const { inboundSymbol, inboundFillAmount, outboundSymbol, outboundFillAmount, takerAddress } = order
   const [ expectedSymbol, actualSymbol, expectedAmount, actualAmount ] = [ inboundSymbol, symbol, inboundFillAmount, amount ]
 
+  logger.debug(`Found routing entry for swap ${swapHash}`)
```
dannypaz: should we log `order` or is that too much information for this log?

treygriffith: I think that's too much information

PR #338, opened 2018-12-20
remove lnd-engine dependency in cli

martineehrlich:
## Description
broker-cli now has a dependency on lnd-engine and we do not have lnd-engine in the package.json nor do we want to just to use it for a status. 
This PR moves the call to lnd-engine to a constant.
This was not noticed because it is not an issue when you use broker-cli locally from the broker directory, but it is when you are using remote cli

## Related PRs

## Todos
- [x] Tests
- [x] Documentation
- [ ] Link to Trello

### DIFF
--- a/broker-cli/commands/health-check.js
+++ b/broker-cli/commands/health-check.js
@@ -15,6 +15,8 @@ const STATUS_CODES = Object.freeze({
 })
 
 /**
+ * Engine status codes we expect to be returned from the engine, defined in:
+ * https://github.com/sparkswap/lnd-engine/blob/master/src/constants/engine-statuses.js
  * @constant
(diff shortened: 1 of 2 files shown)


broker-cli/commands/health-check.js
```diff
@@ -15,6 +14,15 @@ const STATUS_CODES = Object.freeze({
   UNKNOWN: 'UNKNOWN'
 })
 
+/**
+ * @constant
+ * @type {Object<key, String>}
+ * @default
+ */
+const ENGINE_STATUS_CODES = Object.freeze({
```
treygriffith: add a note about where these are defined in LND engine

PR #342, opened 2018-12-27
Feature - watchMarket exponential backoff

### LINKED ISSUE #319: Add exponential backoff to watchMarket
When we start an orderbook by connecting to the Relayer, we handle failures via a `setTimeout` (see: https://github.com/sparkswap/broker/blob/master/broker-daemon/orderbook/index.js#L103).

Instead, we should use an exponential backoff when we encounter successive failures.

bradleystachurski:
## Description
Fixes #319

When an `Orderbook` receives an `end` event from the `watcher`, retries to sync are currently handled with a static retry interval of 5 seconds. This PR updates this logic to exponentially backoff time periods between retries.

It's worth noting that there's already an `exponentialBackoff` function in both the `broker` and `lnd-engine` repos, however this function is implemented to handle retries on initial calls to the `callFunction` and doesn't work for handling `end` events that may be emitted well after the initial call to `watchMarket`.

If the maintainers would prefer a different approach, I'm happy to make changes.

## Related PRs
https://github.com/sparkswap/lnd-engine/pull/110

## Todos
- [x] Tests
- [x] Documentation
- [x] Link to ~~Trello~~ GH Issue


broker-daemon/orderbook/index.js
```diff
@@ -38,6 +38,9 @@ class Orderbook {
     this.bidIndex = new BidIndex(this.store)
     this.logger = logger
     this.synced = false
+    this._retries = 0
```
treygriffith: Can you move the retry logic into its own function or object so as not to muddy the state of the orderbook more generally? It seems like the exponential backoff is independent enough to not need to be mixed in with the other state.

broker-daemon/orderbook/index.js
```diff
@@ -97,12 +101,12 @@ class Orderbook {
       this.synced = false
       watcher.removeListener('sync', onWatcherSync)
       watcher.removeListener('error', onWatcherError)
+      this._updateExponentialDelayPeriod()
```
treygriffith: I'd prefer that a function like this return a value rather than impacting state as a side effect

broker-daemon/orderbook/index.js
```diff
@@ -337,6 +341,39 @@ class Orderbook {
...
+   * @private
+   * @return {void}
+   */
+  _updateExponentialDelayPeriod () {
+    if (this._reachedMaxInterval) {
+      return
+    }
+
+    this._retries++
+    const newDelay = (2 ** this._retries - 1) * 1000
```
treygriffith: some commentary on what you're doing here and why would be helpful

broker-daemon/orderbook/index.js
```diff
@@ -337,6 +341,39 @@ class Orderbook {
...
+
+  /**
+   * Updates retry count and calculates next delay period if the
+   * MAX_RETRY_INTERVAL hasn't been surpassed
+   *
+   * @private
+   * @return {void}
+   */
+  _updateExponentialDelayPeriod () {
+    if (this._reachedMaxInterval) {
```
treygriffith: this seems like an unnecessary state. Can't you just check if the `_delay` is equivalent to the `MAX_RETRY_INTERVAL`? Or just skip this check entirely and let the max logic do the work for you?

broker-daemon/orderbook/index.js
```diff
@@ -337,6 +341,39 @@ class Orderbook {
...
+      return
+    }
+
+    this._retries++
+    const newDelay = (2 ** this._retries - 1) * 1000
+    if (newDelay > MAX_RETRY_INTERVAL) {
```
treygriffith: why not do something like `this._delay = Math.min(newDelay, MAX_RETRY_INTERVAL)` ?

broker-daemon/orderbook/index.spec.js
```diff
@@ -312,9 +319,26 @@ describe('Orderbook', () => {
         expect(watcher.removeListener).to.have.been.calledWith('error', sinon.match.func)
       })
 
-      it('sets a 5 second timeout', () => {
+      it('exponentially backs off after repeated end events', () => {
```
treygriffith: you should add some tests exercising each of the methods that you added

broker-daemon/orderbook/index.js
```diff
@@ -337,6 +341,39 @@ class Orderbook {
...
+      this._delay = newDelay
+    }
+  }
+
+  /**
+   * Resets exponential backoff parameters to initial state
+   * @private
+   * @return {void}
+   */
+  _resetBackoff () {
```
treygriffith: small stylistic note: we're shying away from the `_` prefix as "private" indicator, instead preferring either truly private methods (i.e. not exposed on the object) or just marked as `@private` via jsdoc.

broker-daemon/orderbook/index.js
bradleystachurski: I don't think it's necessary to have the additional methods after refactoring the retry logic, so this is no longer impacting state as a side effect.

Question: Do you want to also implement exponential backoff within `onWatcherError`? If so, I think it makes sense to bring back that method and make it a pure function.

broker-daemon/orderbook/index.js
bradleystachurski: Added a few lines. Let me know if it's too verbose and also if you prefer a different backoff algorithm.

broker-daemon/orderbook/index.spec.js
bradleystachurski: Those additional methods have been removed after refactoring.

I'm not stoked about being unable to stub the new `watchMarketCall` due to its scope, so re-initialization is now tested by resetting the stub history for `watcher.once` and making sure it sets new listeners after the timeout. Let me know if you see a better approach to testing this.

broker-daemon/orderbook/index.js
bradleystachurski: Gotcha, since I'm new to Node culture I'm curious if this is a preference of the team or that's idiomatic JavaScript? Googling around seems to have mixed results.

treygriffith: Consensus is mixed. We use [StandardJS](https://standardjs.com) which is silent on the topic. We've decided as a team to stay away from the `_` prefix (although there are still a couple of places where it exists in the codebase).

treygriffith:
I think I like this refactor, although the multiple nested functions can be a little hard to follow. What do you think @dannypaz ?

broker-daemon/orderbook/index.js
```diff
@@ -71,60 +71,73 @@ class Orderbook {
    * @return {Promise} Resolves when market is being watched (not necessarily when it is synced)
    */
   async watchMarket () {
-    this.logger.debug(`Watching market ${this.marketName}...`)
+    let retries = 0
 
-    const { baseSymbol, counterSymbol } = this
-    const { lastUpdated, sequence } = await this.lastUpdate()
-    const params = { baseSymbol, counterSymbol, lastUpdated, sequence }
+    const watchMarketCall = async () => {
```
martineehrlich: I am not sure this whole thing needs to be wrapped in an anonymous function. It makes it pretty hard to follow. It seems like the anonymous function is being used solely for the purpose of retry logic (unless there is something I am missing). If this is the case, we can pass the retries in as a parameter to watchMarket in the setTimeout call below. (We can default it to 0). :)

broker-daemon/orderbook/index.js
treygriffith: I don't think we need backoff in `onWatcherError` just yet - the errors that trigger that are local (rather than remote) so it's unlikely that we'll need to back off of them.

broker-daemon/orderbook/index.js
bradleystachurski: Hi @martineehrlich, thanks for the feedback!

I agree, the anonymous function makes it hard to follow. I was trying to follow the style used in `lnd-engine` where `validateEngine` uses an anonymous function to handle retries (https://github.com/sparkswap/lnd-engine/blob/628ad14c3448eb2899cc09d796ee93d3e591ea7c/src/index.js#L79-L82), but my goodness is that much cleaner logic than the anonymous function used in `watchMarket`.

Thanks for the suggested changes. I've made the changes and think the logic is _much_ easier to follow. Please let me know if there are any other changes you'd like to see before merging.

broker-daemon/orderbook/index.js
```diff
@@ -98,11 +100,17 @@ class Orderbook {
...
+      // defined in https://en.wikipedia.org/wiki/Exponential_backoff
+      //
+      // Note: Repeated retransmission wait time is typically defined as a random variable between
+      // 0 and 2^c - 1, where c is the number of retries, however here we increase delay time using
+      // 2^c - 1 until MAX_RETRY_INTERVAL is reached.
+      retries++
+      const delay = Math.min((2 ** retries - 1) * 1000, MAX_RETRY_INTERVAL)
+
+      this.logger.info(`Market ${this.marketName} unavailable, retrying in ${delay}ms`, { error })
+      setTimeout(this.watchMarket, delay, retries)
```
treygriffith: will `this.watchMarket` get called in the context of `this`? If not you may need to use a fat arrow function:

```
setTimeout(() => this.watchMarket(retries), delay)
```

broker-daemon/orderbook/index.js
```diff
@@ -85,6 +86,7 @@ class Orderbook {
      */
     const onWatcherSync = () => {
       this.synced = true
+      retries = 0
```
treygriffith: do you have a test for this behavior? Sorry if I missed it.

broker-daemon/orderbook/index.js
treygriffith: Yep, I just tried this branch locally and got:

```
sparkswapd_1  | error: Relayer watchMarket grpc failed {"code":14,"metadata":{"_internal_repr":{}},"details":"EOF","timestamp":"2019-01-08T19:34:21.914Z"}
sparkswapd_1  | info: Market BTC/LTC unavailable, retrying in 1000ms {"error":{"code":14,"metadata":{"_internal_repr":{}},"details":"EOF"},"timestamp":"2019-01-08T19:34:21.915Z"}
sparkswapd_1  | (node:100) UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'debug' of undefined
sparkswapd_1  |     at Timeout.watchMarket [as _onTimeout] (/home/app/broker-daemon/orderbook/index.js:75:17)
sparkswapd_1  |     at ontimeout (timers.js:502:15)
sparkswapd_1  |     at tryOnTimeout (timers.js:323:5)
sparkswapd_1  |     at Timer.listOnTimeout (timers.js:290:5)
sparkswapd_1  | (node:100) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
```

which is due to `this.logger` not being defined in the given context.

broker-daemon/orderbook/index.js
bradleystachurski: Ah good catch! I've added a test for this behavior, however it requires emitting multiple `end` events along with a `sync` event and checking the argument passed to `watchMarket` to get the retry count. If you see a more direct way of testing this behavior I'm happy to update.

broker-daemon/orderbook/index.spec.js
```diff
@@ -312,17 +313,88 @@ describe('Orderbook', () => {
...
+        for (let i = 0; i < 2; i++) {
+          onEnd()
+          const expected = 60000
+          expect(timeoutStub.lastCall).to.have.been.calledWith(sinon.match.func, expected)
+        }
       })
 
       it('re-initializes after the timeout', () => {
         const timeoutFunc = timeoutStub.args[0][0]
+        const updatedRetries = timeoutStub.args[0][2]
```
treygriffith: what is this parameter? I don't think there is a 3rd parameter passed to `setTimeout` anymore, is there?

PR #343, opened 2018-12-28
Add RPC user and pass as environment variables

treygriffith:
## Description
This change sets the RPC user and password in the `.env` file so that they can be updated by a user running `sparkswapd`. It also updates the sample config so that if a user copies the sample config the default authentication will work.

### DIFF
--- a/broker-cli/sample-config.js
+++ b/broker-cli/sample-config.js
@@ -14,30 +14,30 @@ module.exports = {
    * Address of the host for the Broker Daemon gRPC Server
    * @type {String}
    */
-  // rpcAddress: 'localhost:27492',
+  rpcAddress: 'localhost:27492',
 
   /**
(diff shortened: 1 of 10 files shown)


broker-cli/sample-config.js
```diff
@@ -14,30 +14,30 @@ module.exports = {
...
-  // rpcCertPath: '~/.sparkswap/certs/broker-rpc-tls.cert',
+  rpcCertPath: '~/.sparkswap/certs/broker-rpc-tls.cert',
 
   /**
    * Configuration for SSL between the CLI and Daemon. This setting is only required
    * if you will be hosting the daemon remotely
    * @type {Boolean}
    */
-  // disableAuth: false
+  disableAuth: false
```
dannypaz: missing a comma here

PR #345, opened 2018-12-30
move broker to rc6

treygriffith:
## Description
Update version to Release Candidate 6.


treygriffith:
Merging without review since this is just a version bump.

PR #349, opened 2019-01-04
add env setup script

martineehrlich:
## Description
When users setup their env file, we used to have them edit it manually. Once we started having username/password combinations that were supposed to be unique, this became a longer process.

This script automates the setting of the env variables.

## Related PRs

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Link to Trello

### DIFF
--- a/scripts/env-setup.sh
+++ b/scripts/env-setup.sh
@@ -1,0 +1,35 @@
+#!/usr/bin/env bash
+
+# Set the network in the .env file based on user input
+echo "Enter the network: 1 for MainNet, 2 for TestNet"
+read ADDR
+if [ $ADDR = '1' ]; then
+  sed -i "" -e "s/.*NETWORK.*/NETWORK=mainnet/" .env
+elif [ $ADDR = '2' ]; then
+  sed -i "" -e "/NETWORK/c\NETWORK=testnet" .env
+else
+  echo "$ADDR is not a valid option for network"
+  exit 1
+fi
+
+# Set the external btc/ltc addresses in the .env file based on user public IP address
+echo "Enter your public IP address:"
+read IP_ADDRESS
+
+sed -i "" -e "s/^EXTERNAL_BTC_ADDRESS.*/EXTERNAL_BTC_ADDRESS=$IP_ADDRESS/" .env
+sed -i "" -e "s/^EXTERNAL_LTC_ADDRESS.*/EXTERNAL_LTC_ADDRESS=$IP_ADDRESS/" .env
+
+array=(BTC_RPC_USER
+BTC_RPC_PASS
+LTC_RPC_USER
+LTC_RPC_PASS
+RPC_USER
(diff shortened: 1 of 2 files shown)


scripts/env-setup.sh
```diff
@@ -0,0 +1,40 @@
+#!/usr/bin/env bash
+
+################################################
+# Generates and sets environment variables for broker
+#
+################################################
+
+# Set the network in the .env file based on user input
+echo "Enter the network: 1 for MainNet, 2 for TestNet"
```
treygriffith: hmm, I think I'd prefer letters (e.g. `m` for mainnet, `t` for testnet) over numbers. Why did you choose to do it this way?

Also, might be nice to display them in a list, e.g.

```
Enter the network:
m - MainNet
t - TestNet
```

scripts/env-setup.sh
```diff
@@ -0,0 +1,40 @@
...
+read ADDR
```
treygriffith: Why is the variable name here `ADDR`?

scripts/env-setup.sh
```diff
@@ -0,0 +1,40 @@
...
+read IP_ADDRESS
+
+sed -i "" -e "s/^EXTERNAL_BTC_ADDRESS.*/EXTERNAL_BTC_ADDRESS=$IP_ADDRESS/" .env
+sed -i "" -e "s/^EXTERNAL_LTC_ADDRESS.*/EXTERNAL_LTC_ADDRESS=$IP_ADDRESS/" .env
+
+array=(BTC_RPC_USER
+BTC_RPC_PASS
+LTC_RPC_USER
+LTC_RPC_PASS
+RPC_USER
```
treygriffith: do we want to randomly generate the RPC user and pass as well?

How do we expect the user to get those values into their CLI config?

scripts/env-setup.sh
```diff
@@ -0,0 +1,40 @@
...
+################################################
+# Generates and sets environment variables for broker
+#
+################################################
+
+# Set the network in the .env file based on user input
+echo "Enter the network: 1 for MainNet, 2 for TestNet"
+read ADDR
+if [ $ADDR = '1' ]; then
+  sed -i "" -e "s/.*NETWORK.*/NETWORK=mainnet/" .env
```
treygriffith: do we want to first copy over a sample .env file? Or should that happen somewhere else? Or maybe that only happens if the .env file doesn't exist?

martineehrlich: We have them do it manually but i guess we could do it based on the network they choose 🤷‍♀️

scripts/env-setup.sh
martineehrlich: We tell them to get it from their .env file in the instructions. It seemed less complicated to do it this way, than to have them manually add anything to the .env file. Thoughts?

PR #354, opened 2019-01-04
Protofile standard changes

dannypaz:
## Description
This PR fixes an issue with our grpc proto file that prevents `protoc` from compiling in other languages.

We receive a warning stating that `google.api.http` is being resolved on additional tries and that `google.api.http` is not available at the parent level. This is fixed by adding a `.` at the begin of the import.

There were additional changes that needed to be made, specifically that protoc will yell at us that the import in `annotations.proto` for `http.proto` is incorrect. However, until we upgrade gRPC on the broker, we are not able to change this path (protobufjs upgrade which includes a `root` flag)

Additional Changes:
1. added 404 handling to grpc gateway
2. removed unused flags on grpc gateway (`grpc.load` does not take an object)
3. added jsdoc

## Related PRs

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Link to Trello

### DIFF
--- a/broker-daemon/utils/grpc-gateway.js
+++ b/broker-daemon/utils/grpc-gateway.js
@@ -1,19 +1,58 @@
 /**
- * @author konsumer
- * @see {@link https://github.com/konsumer/grpc-dynamic-gateway}
+ * Sparkswap GRPC Gateway HTTP Proxy
  *
+ * This loads a proto file and creates REST endpoints from it's google.api.http declarations.
+ *
+ * We 'forked' this library from the origin author konsumer which can be found at
+ * the link below.
+ *
+ * @see {@link https://github.com/konsumer/grpc-dynamic-gateway}
  */
 
-const requiredGrpc = require('grpc')
+const grpc = require('grpc')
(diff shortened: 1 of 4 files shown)


dannypaz:
[protoc fails when processing broker.proto](https://trello.com/c/0t03qN6W/1028-protoc-fails-when-processing-brokerproto)

broker-daemon/utils/grpc-gateway.js
```diff
@@ -1,43 +1,73 @@
...
-  if (include.endsWith('/')) {
-    include = include.substring(0, include.length - 1) // remove"/"
-  }
-  protoFiles = protoFiles.map(function (value, index, array) {
-    if (value.startsWith(include)) {
-      value = value.substring(include.length + 1)
-    }
-    return value
-  })
-  const protos = protoFiles.map(p => include ? grpc.load({ file: p, root: include }) : grpc.load(p))
```
dannypaz: I removed the lines w/ `include` because our version of grpc.load does not take an object (it will fail).

PR #356, opened 2019-01-06
Fix: add cancelled status for underlying orders

treygriffith:
## Description
We have a cancelled status for underlying orders, but it was missing from the proto file. This adds it back to the proto file to avoid errors with the CLI and other clients.


dannypaz:
@treygriffith we should remove the version bump from this PR since we'll have a few other changes we'll want to include

PR #357, opened 2019-01-06
Fix: remove listeners when an order is cancelled

treygriffith:
## Description
The broker has a bug wherein if an order is cancelled, the Broker will cancel it, but will continue listening for updates from the Relayer.

Beyond creating a memory leak, this creates an issue where the Relayer closing the stream (which it will do after cancelling an order) causes the Broker to move the order to an errored state. This is particularly problematic if the order has not yet been fully transitioned to a cancelled state, as it will cause an invalid transition (moving to rejected while still in the cancel transition), crashing the broker.

## Todos
- [x] Tests

### DIFF
--- a/broker-daemon/state-machines/order-state-machine.js
+++ b/broker-daemon/state-machines/order-state-machine.js
@@ -289,30 +289,31 @@ const OrderStateMachine = StateMachine.factory({
       }
       const dataHandler = ({ orderStatus, fill }) => {
         try {
-          // the Relayer will send a single data message containing the order's state as cancelled and close
-          // the stream if the order has been cancelled. We should handle that and cancel the order locally.
           if (OrderStateMachine.STATES[orderStatus] === OrderStateMachine.STATES.CANCELLED) {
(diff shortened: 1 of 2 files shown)


broker-daemon/state-machines/order-state-machine.js
```diff
@@ -289,30 +289,31 @@ const OrderStateMachine = StateMachine.factory({
       }
       const dataHandler = ({ orderStatus, fill }) => {
         try {
-          // the Relayer will send a single data message containing the order's state as cancelled and close
-          // the stream if the order has been cancelled. We should handle that and cancel the order locally.
           if (OrderStateMachine.STATES[orderStatus] === OrderStateMachine.STATES.CANCELLED) {
-            this.logger.info(`Order ${orderId} was cancelled on the relayer, cancelling locally.`)
-            return this.tryTo('cancel')
```
treygriffith: this is the offending line as it prevents `finish()` from being called, which tears down the listeners.

PR #371, opened 2019-01-11
remove auto complete information

dannypaz:
## Description
This PR removes text explaining auto complete setup for the broker-cli due to the following CLI issues:

1. what using auto-complete with the command `sparkswap`, all sub commands will have an escaped `/` space between them. ex: `sparkswap wallet/ balance`
2. when using auto-complete inside of a command like `sparkswap wallet`, all commands will be duplicated and contain the same issue as #1. ex: `sparkswap wallet wallet/ balance`

Both of these are caused by our solution for sub-modules in caporal AND how `tabtab` (a dep of caporal) parses the `name` of the command. Caporal does not support sub-modules for our use.

Example of error:
screen shot 2019-01-11 at 1 21 35 pm

## Related PRs

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Link to Trello


dannypaz:
[auto-completion adds a slash in subcommands](https://trello.com/c/B4k8DAcs/998-auto-completion-adds-a-slash-in-subcommands)

PR #373, opened 2019-01-12
Break up orders to meet the max payment size

treygriffith:
## Description
This PR acknowledges the maximum payment size imposed by the payment channel network engine and breaks up placed orders into multiple smaller placed orders, each under the max size.

## Questions
- Should this maximum size be set by the Relayer rather than by the engine?
  - **Answer:** not right now, but we may want to revisit.
- Should this maximum be set in a way to leave room for routing fees? (and if so, can that be combined with the above to push responsibility for calculating the maximum to the Relayer?)
  - **Answer:** LND only uses the payment intent rather than the total of the route, so this maximum should be fine.

## Related PRs
https://github.com/sparkswap/lnd-engine/pull/148 is required (and must be referenced) for this to work, as it uses `maxPaymentSize` from the engine configuration.

## Todos
- [x] Tests
- [x] Documentation

## Future considerations
This change may set the stage for breaking up orders to take advantage of multiple channels and allow us to support more flexible increase and decrease of committed funds prior to splicing.


broker-daemon/block-order-worker/index.js
```diff
@@ -557,6 +557,62 @@ class BlockOrderWorker extends EventEmitter {
...
+    if (Big(quantumPrice).gte(maxPaymentSizeImpliedPrice)) {
+      // counter for the block order is larger than base (relative to their max payment sizes)
+      maxBaseAmountPerOrder = Big(counterMaxPayment).div(quantumPrice).round(0)
+    } else {
+      // base for the block order is larger than counter (relative to their max payment sizes)
+      maxBaseAmountPerOrder = Big(baseMaxPayment)
+    }
+
+    let baseAmountRemaining = Big(baseAmount)
+    let orderCount = 1
```
dannypaz: this never gets incremented? Also since _place_orders can be called multiple times for a blockorder, there is the potential to have a bunch of logs that say `Placing order 1` which might not be what you were looking for?

broker-daemon/block-order-worker/index.js
```diff
@@ -557,6 +557,62 @@ class BlockOrderWorker extends EventEmitter {
...
   }
 
+  /**
+   * Place orders for a block order for a given amount, breaking them up based on the maximum order size
+   * @param  {BlockOrder} blockOrder Block Order to place orders on behalf of
+   * @param  {String}     baseAmount Int64 amount, in the base currency's base units, to place orders for
+   * @return {void}
+   */
+  _placeOrders (blockOrder, baseAmount) {
+    // order params
```
dannypaz: we dont need this log

broker-daemon/block-order-worker/index.js
```diff
@@ -557,6 +557,62 @@ class BlockOrderWorker extends EventEmitter {
...
+    let orderCount = 1
+
+    // split our larger block order into individual placed orders that are each under
+    // the max payment size
+    while (baseAmountRemaining.gt(0)) {
+      this.logger.info(`Placing order #${orderCount} for BlockOrder`, { blockOrderId: blockOrder.id })
+
+      let orderBaseAmount = baseAmountRemaining
+
+      if (orderBaseAmount.gte(maxBaseAmountPerOrder)) {
```
dannypaz: When we are placing orders for a particular block order, do you think it matters whether we have an even amount for orders?

Example situation: 
1. max amount per order is 200
2. we have an order placed for 500

you would have the following order scheme:
`200, 200, 100`

Instead, we could have a formula to split the current order into a particular size:
`167, 167, 166`

Is this even a thing?

treygriffith: That might be an optimization to do later, but I don't think it impacts anything right now

broker-daemon/block-order-worker/index.js
treygriffith: Yep I missed the `++`. Fixing now.

`_placeOrders` can be called more than once, but should be called only once per `workBlockOrder`, so I still think this log is useful - you should have conflicting logs at the same moment.

PR #377, opened 2019-01-17
Returning JSON response for Broker CLI.

mdasif:
## Description
Adds option `--json` to print a json response for all commands. 

It also adds an option `--no-stream` which will print current snapshot of orderbook output. 

## Related PRs

## Todos
- [x] Tests
- [ ] Documentation
- [x] Link to Trello


broker-cli/commands/health-check.js
```diff
@@ -26,6 +26,92 @@ const ENGINE_STATUS_CODES = Object.freeze({
...
+/**
+ * Prints health check status summary in JSON format.
+ * @param {Object} engineStatus
+ * @param {Object} orderbookStatus
+ * @param {Object} relayerStatus
+
+ * @returns {Void}
+ */
+
+function createHealthCheckJson (engineStatus, orderbookStatus, relayerStatus) {
```
treygriffith: this change looks like it needs tests, unless I missed them

broker-cli/commands/health-check.js
```diff
@@ -26,6 +26,92 @@ const ENGINE_STATUS_CODES = Object.freeze({
...
+  const statuses = []
+
+  if (engineStatus.length > 0) {
+    engineStatus.forEach(({ symbol, status }) => {
+      const statusString = (status === ENGINE_STATUS_CODES.VALIDATED) ? `${STATUS_CODES.OK}`.green : status.red
```
treygriffith: we shouldn't be colorizing output that is used as JSON as colors will mess with the output text and potentially make it unusable for scripting

broker-cli/commands/health-check.js
```diff
@@ -26,6 +26,92 @@ const ENGINE_STATUS_CODES = Object.freeze({
...
+/**
+ * Prints health check status summary in Table format.
+ * @param {Object} engineStatus
+ * @param {Object} orderbookStatus
+ * @param {Object} relayerStatus
+
+ * @returns {Void}
+ */
+
+function createHealthCheckTable (engineStatus, orderbookStatus, relayerStatus) {
```
treygriffith: is there any way we can refactor this and the json function? it seems like a lot of duplicated code

broker-cli/utils/strings.js
```diff
@@ -1,7 +1,9 @@
 const RPC_ADDRESS_HELP_STRING = 'Location of the RPC server to use'
 const MARKET_NAME_HELP_STRING = 'Relevant market name'
+const JSON_FORMAT_STRING = 'Print in JSON format'
```
treygriffith: Perhaps "Print output in JSON format" would be more descriptive

broker-cli/commands/order/summary.js
```diff
@@ -37,6 +37,30 @@ function createUI (market, orders) {
...
 }
 
+/**
+ * Prints the users orders in JSON format
+ * @param {String} market
+ * @param {Array.<{blockOrderId, side, amount, price, timeInForce, status>}}
+
+ * @returns {Void}
+ */
+function createSummaryJson (market, orders) {
```
treygriffith: This needs tests (unless I missed them)

broker-cli/commands/order/summary.js
```diff
@@ -37,6 +37,30 @@ function createUI (market, orders) {
...
+  const orderList = orders.map((order) => {
+    const price = order.isMarketOrder ? 'MARKET' : order.limitPrice
+    const side = order.side === ORDER_TYPES.BID ? order.side.green : order.side.red
```
treygriffith: same comment about colors

broker-cli/commands/orderbook.js
```diff
@@ -95,6 +95,40 @@ function createUI (market, asks, bids) {
...
+function createOrderBookJson (market, asks, bids) {
+  console.clear()
+  const askList = asks.map((ask) => {
+    return {
+      price: ask.price,
+      amount: ask.amount
+    }
+  })
+
+  const bidList = bids.map((bid) => {
```
treygriffith: Is this map doing anything? Are there properties that it is filtering out?

broker-cli/commands/orderbook.js
```diff
@@ -178,5 +224,6 @@ module.exports = (program) => {
     .command('orderbook', 'View the order book for a specific market')
     .option('--market <marketName>', MARKET_NAME_HELP_STRING, validations.isMarketName, null, true)
     .option('--rpc-address [rpc-address]', RPC_ADDRESS_HELP_STRING, validations.isHost)
+    .option('--json', JSON_FORMAT_STRING, program.BOOLEAN)
```
treygriffith: It doesn't make sense to have a json output of a streaming function. If using json, we should perhaps either assume, or require that the user use the non-streaming endpoint `GetOrderbook`

broker-cli/commands/wallet.js
```diff
@@ -46,6 +46,72 @@ const SUPPORTED_COMMANDS = Object.freeze({
...
+    let {
+      symbol,
+      error = null,
+      totalChannelBalance,
+      totalPendingChannelBalance,
+      uncommittedBalance,
+      uncommittedPendingBalance
+    } = balance
+
+    totalChannelBalance = error ? 'Not Available' : totalChannelBalance
```
treygriffith: A lot of these types of things that we do for display in the CLI we can probably skip when doing JSON display

broker-cli/commands/wallet.js
```diff
@@ -46,6 +46,72 @@ const SUPPORTED_COMMANDS = Object.freeze({
...
+    return {
+      symbol: symbol,
+      totalChannelBalance: totalChannelBalance,
+      totalPendingChannelBalance: totalPendingChannelBalance,
+      uncommittedBalance: uncommittedBalance,
+      uncommittedPendingBalance: uncommittedPendingBalance
+    }
+  })
+
+  logger.info({ 'walletBalances': balancesArray })
```
treygriffith: why are we using `logger.info` here when other places we use `console.log`?

treygriffith: why is `walletBalances` in quotes?
Also, how did you choose `walletBalances` as the name here?

broker-cli/commands/wallet.js
```diff
@@ -290,6 +329,111 @@ const CAPACITY_STATUSES = Object.freeze({
...
+        receiveCapacity: formatBalance(baseSymbolCapacities.pendingReceiveCapacity, NETWORK_STATUSES.AVAILABLE),
+        sendCapacity: formatBalance(baseSymbolCapacities.outstandingSendCapacity, NETWORK_STATUSES.AVAILABLE)
+      },
+      {
+        type: 'inactive',
+        receiveCapacity: formatBalance(baseSymbolCapacities.inactiveReceiveCapacity, NETWORK_STATUSES.AVAILABLE),
+        sendCapacity: formatBalance(baseSymbolCapacities.inactiveSendCapacity, NETWORK_STATUSES.AVAILABLE)
+      }
+    ],
+    baseSymbolCapacitiesStatus: baseSymbolCapacities.status,
```
treygriffith: what is this value?

broker-cli/commands/wallet.js
```diff
@@ -290,6 +329,111 @@ const CAPACITY_STATUSES = Object.freeze({
...
+  // If any balances in the following table are empty (which occurs if the engine is unavailable)
+  // then the text will show up as `Not Available`.
+  //
+  // The user will then be prompted with a warning letting them know that we failed
+  // to receive a balance for a particular currency
+  const networkStatus = {
+    market: market,
+    baseSymbol: baseSymbol,
+    counterSymbol: counterSymbol,
+    statuses: [
```
treygriffith: I'm not sure that an array is the best format for this, a straight object might be better, e.g.

```
...
statuses: {
  available: {
    sendCapacity:
...
```

Or better yet, why not format it more like the actual return value of the RPC?

treygriffith:
@mdasif After thinking about it a bit more, it seems like the best way to deliver JSON to the user would actually be to return the default JSON value produced by protobuf/grpc. That way we don't need to invent or document yet another serialization format for the same data.

The serialization shown in the "javascript" tab of https://sparkswap.com/docs/broker/api/ is probably what I would expect if I stuck `--json` on the end of any given command.

Structuring it that way would also reduce the size of this PR, simplify it and its testing.

dannypaz:
@mdasif an example of what @treygriffith is talking about https://github.com/sparkswap/broker/pull/377#issuecomment-455428518

Inside of `healthcheck.js`, instead of breaking apart the table and formatting into different functions, we can do the following:

```
    const res = await client.adminService.healthCheck({})

    if (json) {
      console.log(JSON.stringify(res))
      return
    }

    const {
      engineStatus = [],
      orderbookStatus = [],
      relayerStatus = STATUS_CODES.UNKNOWN
    } = res
```

For every call, we can get the response object and IF we want to return JSON, we simply check the `json` flag and `return` early from the command (without additional processing)

dannypaz:
@mdasif this PR failed on formatting issues. run `npm run format` to see the errors

treygriffith:
@mdasif which commands are still to be done? Are they intended to be done as part of this PR?

mdasif:
@treygriffith this adds support for json for all the PRs which have a table output. `wallet`, `health-check` and `order summary`.

treygriffith:
@mdasif is this the entire scope of this PR with `--json` flags being added to other commands later, or is this PR still work-in-progress?

mdasif:
Added `--json` option for all the commands except `orderbook`, All the commands at present just prints the grpc response so json flag will have no effect on the output but it will be a valid option for all the commands to consistency.

broker-cli/commands/orderbook.js
```diff
@@ -107,11 +107,18 @@ function createUI (market, asks, bids) {
...
  */
 async function orderbook (args, opts, logger) {
-  const { market, rpcAddress } = opts
+  const { market, rpcAddress, json } = opts
   const request = { market }
 
   try {
     const brokerDaemonClient = new BrokerDaemonClient(rpcAddress)
+
+    if (json) {
```
treygriffith: Per our conversation, I think there should be an option for a static orderbook that still displays the UI (#387), and I would consider the json option only valid with that option as well. But for this PR I think it's ok to merge.

dannypaz:
@mdasif @treygriffith what APIs are we adding JSON too? Can we list these in the description?

mdasif:
@dannypaz updated description to include the commands impacted by this change and using `JSON.stringify` to convert response into json.

broker-cli/commands/orderbook.js
```diff
@@ -107,11 +106,24 @@ function createUI (market, asks, bids) {
...
  */
 async function orderbook (args, opts, logger) {
-  const { market, rpcAddress } = opts
+  const { market, rpcAddress, json, stream } = opts
   const request = { market }
 
   try {
     const brokerDaemonClient = new BrokerDaemonClient(rpcAddress)
+
+    if (!stream) {
```
treygriffith: if the client uses `--json` but not `--stream` we should show an error, otherwise they will pass `--json` but it will have no effect.

I also think that streaming should be the default, so the option is better as `--no-stream` or `--static`

broker-cli/commands/buy.spec.js
```diff
@@ -90,4 +89,13 @@ describe('buy', () => {
     expect(createBlockOrderSpy).to.have.been.called()
     expect(createBlockOrderSpy).to.have.been.calledWith(expectedRequest)
   })
+
+  describe('with json output', () => {
+    it('logs buy result', async () => {
+      const json = true
+      opts = { market, timeInForce, rpcAddress, json }
+      await buy(args, opts, logger)
+      expect(infoSpy).to.have.been.calledOnce()
```
treygriffith: should test what this is being called with too, right?

broker-cli/commands/buy.js
```diff
@@ -40,7 +40,7 @@ async function buy (args, opts, logger) {
   try {
     const client = new BrokerDaemonClient(rpcAddress)
     const blockOrderResult = await client.orderService.createBlockOrder(request)
-    logger.info(blockOrderResult)
+    logger.info(JSON.stringify(blockOrderResult))
```
treygriffith: do we need to call `JSON.strinfigy` here? What's the behavior if we dont?

broker-cli/commands/order/summary.js
```diff
@@ -49,12 +49,17 @@ function createUI (market, orders) {
...
-  const { market, rpcAddress } = opts
+  const { market, rpcAddress, json } = opts
   const request = { market }
   try {
     const brokerDaemonClient = new BrokerDaemonClient(rpcAddress)
     const orders = await brokerDaemonClient.orderService.getBlockOrders(request)
-    createUI(market, orders.blockOrders)
+
+    if (json) {
+      console.log(JSON.stringify(orders))
```
treygriffith: from a consistency standpoint, in most of the other commands you do an `if...return` but here you do an `if...else`. any reason for the difference? If not, I would say settle on one style and stick with it.

treygriffith:
@mdasif your description of the PR is weird - at first it says it's for selected commands, then for all commands, and finally it lists commands (but the actual changes include commands that aren't listed). Can you clean that up and make the description match the PR?

broker-cli/commands/orderbook.js
mdasif: Using `--no-stream` as option and returning failure if json and is passed without --no-stream

broker-cli/commands/buy.js
dannypaz: @treygriffith I think this is what you mean. I made a comment about this earlier. 

If we simply pass the object (in this case `blockOrderResult`) to `logger` it _wont actually_ be valid json. We need to use `JSON.stringify` on the return values (stdout)

mdasif: I have split the json and non-json output to . not touch the non-json output for now. we can beautify the non-json results separately.

broker-cli/commands/orderbook.spec.js
```diff
@@ -208,7 +208,7 @@ describe('orderbook', () => {
       json = true
       consoleStub = { log: sinon.stub() }
       program.__set__('console', consoleStub)
-      opts = { market, rpcAddress, json }
+      opts = { market, rpcAddress, json, noStream: true }
```
treygriffith: what is `noStream`? It thought the option was `static`?

dannypaz:
[Add `--json` flag to kcli commands](https://trello.com/c/eQTbEZAK/461-add-json-flag-to-kcli-commands)

dannypaz:
Closing so we can create smaller tickets for this issue

PR #378, opened 2019-01-18
FEATURE - sparkswapd default configuration 

dannypaz:
## Description
This is the first PR in the epic for having a standalone Broker docker image. The major piece of this change is adding default values for the broker/rpc/relayer inside of `config.json`. This change will allow us to package all default configuration into the docker image instead of requiring the user to specify a ton of environment variables to get the broker container to work.

Two Additional Benefits:
1. i've cleaned up the CLI parameters for `sparkswapd` (view the screenshot below)
2. `sparkswapd.js` no longer relies on `process.env` (instead environment variables are imported via the command line in `start-sparkswapd.sh`)

*These changes allow for the following workflow:*
1. Setting Environment Variables in `.env` (docker-compose)
2. Environment variables are then used in the `start-sparkswapd.sh` file and are then imported into `sparkswapd` through the use of CLI options
3. Defaults are used correctly in the CLI `sparkswapd.js`

screen shot 2019-01-18 at 10 56 43 am

## Related PRs

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Link to Trello


.env-mainnet-sample
```diff
@@ -1,50 +1,41 @@
...
 RPC_PASS=sparkswap
 
 # Below are the urls that are used for sparkswapd to be connected to a
 # sparkswap hosted relayer.
 #
-# Relayer RPC Hosts for MainNet
-RELAYER_RPC_HOST=relayer.mainnet.sparkswap.com:28492
-RELAYER_CERT_HOST=relayer.mainnet.sparkswap.com:8080
+# Relayer RPC Hosts for TestNet
+RELAYER_RPC_HOST=relayer.testnet.sparkswap.com:28492
```
treygriffith: This seems like the wrong relayer for mainnet?

.env-regtest-sample
```diff
@@ -1,53 +1,34 @@
...
 
 # Set the network that sparkswap and engines will work on.
-# Valid options are simnet, regtest, and testnet.
-#
 NETWORK=regtest
 
-# RPC User and Password for accessing the Sparkswap Daemon. If you are accessing
-# the daemon from outside of a closed network, you should update these values
-# for your security.
-RPC_USER=sparkswap
```
treygriffith: Why remove the the `RPC_USER` and `RPC_PASS` from this file?

broker-daemon/bin/sparkswapd.js
```diff
@@ -5,81 +5,80 @@
...
+const { version: CLI_VERSION } = require('../../package.json')
 
 // TODO: Change this path to be sparkswapd specific
 const { validations } = require('../../broker-cli/utils')
 
-const { version: CLI_VERSION } = require('../../package.json')
+// LND currently uses ECDSA generated certs, which we use on the broker. We've
+// add ECDSA to the gRPC cipher suites as it is not added by default.
+// If this line is removed, we will see an SSL Handshake Error stating `no shared cipher`
+process.env.GRPC_SSL_CIPHER_SUITES = 'HIGH+ECDSA:ECDHE-RSA-AES128-GCM-SHA256:ECDHE-RSA-AES256-GCM-SHA384'
```
treygriffith: did you include this here because this is effectively application code, and not really something that ought to be changed?

broker-daemon/bin/sparkswapd.js
```diff
@@ -5,81 +5,80 @@
...
+  .option('--broker.id-pub-key [id-pub-key]', 'Location of the public key for the broker\'s identity', validations.isFormattedPath, config.broker.idPubKey)
+  .option('--broker.id-priv-key [id-priv-key]', 'Location of private key for the broker\'s identity', validations.isFormattedPath, config.broker.idPrivKey)
+  // TODO: Make this a list instead of string
+  .option('--broker.markets [markets]', 'Comma-separated market names to track on startup', validations.areValidMarketNames, config.broker.markets)
+  .option('--broker.disable-auth [disable-auth]', 'Disable SSL for the broker (DEV ONLY)', program.BOOL, config.broker.disableAuth)
+  // Relayer configuration options
+  .option('--relayer.rpc-host [rpc-host]', 'The host address for the SparkSwap Relayer', validations.isHost, config.relayer.rpcHost)
+  .option('--relayer.cert-path [cert-path]', 'Location of the root certificate for the SparkSwap Relayer (only used in development)', validations.isFormattedPath, config.relayer.certPath)
+  // Broker RPC configuration options
+  .option('--rpc.rpc-address [rpc-rpc-address]', 'Add a host/port to listen for daemon RPC connections', validations.isHost, config.rpc.rpcAddress)
```
treygriffith: why name this `rpc.rpc-address` instead of `rpc.address` or `rpc-address`?

broker-daemon/bin/sparkswapd.js
```diff
@@ -5,81 +5,80 @@
...
   let lowerSymbol = currency.symbol.toLowerCase()
-  let ENGINE_TYPE = process.env[`${currency.symbol}_ENGINE_TYPE`]
-  let LND_TLS_CERT = process.env[`${currency.symbol}_LND_TLS_CERT`]
-  let LND_MACAROON = process.env[`${currency.symbol}_LND_MACAROON`]
-  let LND_RPC_HOST = process.env[`${currency.symbol}_LND_RPC_HOST`]
+
   program
-    .option(`--${lowerSymbol}-engine-type <type>`, `The type of underlying Payment Channel Network node for ${currency.name}`, [ 'LND' ], ENGINE_TYPE)
+    .option(`--${lowerSymbol}.engine-type [engine-type]`, `The type of underlying Payment Channel Network node for ${currency.name}`, isSupportedEngineType, currency.engineType)
+    .option(`--${lowerSymbol}.rpc-host [rpc-host]`, `Location of the LND RPC server to use for ${currency.name}.`, validations.isHost, currency.rpcHost)
```
treygriffith: don't you want to group this under an lnd-specific command, by naming it something like `--${lowerSymbol}.lnd-rpc-host` ?

broker-daemon/config.json
```diff
@@ -1,14 +1,42 @@
 {
+  "broker": {
```
treygriffith: do these need to be grouped under `broker`? or can they be top level?

scripts/start-sparkswapd.sh
```diff
@@ -7,13 +7,72 @@
 #            as `/secure` is a volume inside of the container
 ################################################
 
-set -e
+set -ex
```
treygriffith: do you want to set `-u` as well to protect from unbound variables?

scripts/start-sparkswapd.sh
```diff
@@ -7,13 +7,72 @@
...
+
+# We set value defaults here to mimic the behaivor of docker-compose in-case
+# this script is called outside of docker-compose
+DATA_DIR=${DATA_DIR:-""}
```
treygriffith: We should consider matching up these names with the actual config parameter, e.g. (if we stick with the `broker` prefix) `BROKER_DATA_DIR` instead of `DATA_DIR`

.env-regtest-sample
dannypaz: @treygriffith I removed these because they are set as defaults, however it is probably good to add them back in the sample for consistency

broker-daemon/bin/sparkswapd.js
dannypaz: @treygriffith I removed the lnd prefix because rpc-host should be pretty generic for us. However I should prefix the macaroon tag w/ `lnd`

broker-daemon/config.json
dannypaz: @treygriffith they can be top level

scripts/start-sparkswapd.sh
dannypaz: I went ahead and removed broker prefix so we should be OK

broker-daemon/bin/sparkswapd.js
treygriffith: I might change the description for the generic ones to be more generic then

broker-daemon/bin/sparkswapd.js
```diff
@@ -5,97 +5,104 @@
...
+  .option('--broker.id-priv-key [id-priv-key]', 'Location of private key for the broker\'s identity', validations.isFormattedPath, config.idPrivKey)
+  // TODO: Make this a list instead of string
+  .option('--broker.markets [markets]', 'Comma-separated market names to track on startup', validations.areValidMarketNames, config.markets)
+  .option('--broker.disable-auth [disable-auth]', 'Disable SSL for the broker (DEV ONLY)', program.BOOL, config.disableAuth)
+  // Relayer configuration options
+  .option('--relayer.rpc-host [rpc-host]', 'The host address for the SparkSwap Relayer', validations.isHost, config.relayer.rpcHost)
+  .option('--relayer.cert-path [cert-path]', 'Location of the root certificate for the SparkSwap Relayer (only used in development)', validations.isFormattedPath, config.relayer.certPath)
+  // Broker RPC configuration options
+  .option('--rpc.address [rpc-address]', 'Add a host/port to listen for daemon RPC connections', validations.isHost, config.rpc.rpcAddress)
+  .option('--rpc.rpc-user [rpc-user]', 'Broker rpc user name', program.String, config.rpc.rpcUser)
```
treygriffith: this option is still `--rpc.rpc-user` which is duplicative and inconsistent with `--rpc.address` above

broker-daemon/bin/sparkswapd.js
```diff
@@ -5,97 +5,104 @@
...
-  .option('--rpc-pubkey-path [path]', 'Location of the public key for the broker\'s rpc', validations.isFormattedPath, RPC_PUB_KEY)
-  .option('--disable-relayer-auth [disable-relayer-auth]', 'Disable SSL and message signing to the relayer (DEV ONLY)', program.BOOL, DISABLE_RELAYER_AUTH)
-  .option('--id-privkey-path [path]', 'Location of private key for the broker\'s identity', validations.isFormattedPath, ID_PRIV_KEY)
-  .option('--id-pubkey-path [path]', 'Location of the public key for the broker\'s identity', validations.isFormattedPath, ID_PUB_KEY)
-  .option('--markets [markets]', 'Comma-separated market names to track on startup', validations.areValidMarketNames, MARKETS)
-  .option('--relayer-host [server]', 'The host address for the SparkSwap Relayer', validations.isHost, RELAYER_RPC_HOST)
-  .option('--relayer-cert-path [path]', 'Location of the root certificate for the SparkSwap Relayer', validations.isFormattedPath, RELAYER_CERT_PATH)
+  // Broker configuration options
+  .option('--broker.data-dir [data-dir]', 'Location to store SparkSwap data', validations.isFormattedPath, config.dataDir)
```
treygriffith: if these options are no longer in a `broker` section in the config, I would think they would also be top level options for `sparkswapd`. Also, since `sparkswapd` *is* the broker, it's a bit weird to have options that are prefixed as `broker`. Aren't they all broker options?

dannypaz:
[Create Standalone Broker Docker Image](https://trello.com/c/rL15yr8m/1072-create-standalone-broker-docker-image)

broker-daemon/bin/sparkswapd.js
```diff
@@ -5,97 +5,104 @@
...
-  .option('--relayer-cert-path [path]', 'Location of the root certificate for the SparkSwap Relayer', validations.isFormattedPath, RELAYER_CERT_PATH)
+  // Broker configuration options
+  .option('--data-dir [data-dir]', 'Location to store SparkSwap data', validations.isFormattedPath, config.dataDir)
+  .option('--interchain-router-address [interchain-router-address]', 'Add a host/port to listen for interchain router RPC connections', validations.isHost, config.interchainRouterAddress)
+  .option('--id-pub-key [id-pub-key]', 'Location of the public key for the broker\'s identity', validations.isFormattedPath, config.idPubKey)
+  .option('--id-priv-key [id-priv-key]', 'Location of private key for the broker\'s identity', validations.isFormattedPath, config.idPrivKey)
+  // TODO: Make this a list instead of string
+  .option('--markets [markets]', 'Comma-separated market names to track on startup', validations.areValidMarketNames, config.markets)
+  .option('--disable-auth [disable-auth]', 'Disable SSL for the broker (DEV ONLY)', program.BOOL, config.disableAuth)
```
treygriffith: This doesn't need an actual parameter does it? Is the right way to list this: `'--disable-auth'` instead of including `[disable-auth]`? (I know your PR didn't introduce this, just noticed it).

broker-daemon/bin/sparkswapd.js
```diff
@@ -5,97 +5,104 @@
...
       disableAuth,
-      rpcPrivkeyPath: privRpcKeyPath,
-      rpcPubkeyPath: pubRpcKeyPath,
+      relayerRpcHost,
+      relayerCertPath,
+      rpcAddress,
       rpcUser,
-      rpcPass
+      rpcPass,
+      rpcPubKey: pubRpcKeyPath,
```
treygriffith: These are tough, because you probably want to keep the `path` suffix for the BrokerDaemon options, but at the very least these (and the id ones) would make more sense to me if the options the BrokerDaemon took were `rpcPubKeyPath` and `idPrivKeyPath` rather than the current setup.

PR #253, opened 2018-09-25
add trades command, gettrades rpc, and edit marketevent

martineehrlich:
## Description
Endpoint where to query trades by market. 
- Add trades CLI command
- Add getTrades rpc
- Filter trades by since and limit

## Related PRs
List related PRs if applicable

## Todos
- [x] Tests
- [x] Documentation
- [x] Link to Trello

PR #381, opened 2019-01-20
Feature - add limit to getOrderbook

### LINKED ISSUE #320: Add limit to GetOrderbook RPC
The [`GetOrderbook` RPC](https://github.com/sparkswap/broker/blob/master/broker-daemon/broker-rpc/orderbook-service/get-orderbook.js) currently only takes the `market` as input and returns the entire orderbook for that market.

The `GetOrderbook` RPC should optionally take a parameter to limit the values returned by the RPC, possible a `limit` to limit the total number of orders returned.

bradleystachurski:
## Description
Closes #320

The `GetOrderbook` RPC only takes a `market` parameter. This PR adds an optional `limit` parameter that limits the total number of orders returned for either side of the orderbook.

## Related PRs
https://github.com/sparkswap/broker/pull/253
The above PR tackles a similar implementation of adding a limit parameter to the `getTrades` RPC.

## Todos
- [x] Tests
- [x] Documentation
- [x] Link to ~~Trello~~ GH Issue


broker-daemon/broker-rpc/orderbook-service/get-orderbook.js
```diff
@@ -1,10 +1,20 @@
 const nano = require('nano-seconds')
+const { Big } = require('../../utils')
+
+/**
+ * Default limit for number of orders returned for each side of orderbook
+ * @type {Integer}
```
dannypaz: [nit] would use type `Big` here

broker-daemon/broker-rpc/orderbook-service/get-orderbook.js
```diff
@@ -25,6 +35,9 @@ async function getOrderbook ({ params, logger, orderbooks }, { GetOrderbookRespo
     const datetime = nano.toISOString(currentTime)
     const orders = await orderbook.all()
 
+    // limit is passed as an integer with default protobuf value of '0', so '0' implies limit was omitted
+    const limit = (params.limit === '0' || params.limit === undefined) ? DEFAULT_LIMIT : Big(params.limit)
```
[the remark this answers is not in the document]
bradleystachurski: Good question. Quick answer is no, I wouldn't expect `params.limit` to be undefined.

I defined `limit` to be consistent with the implementation of the `limit` param in `getTrades` (https://github.com/sparkswap/broker/blob/master/broker-daemon/broker-rpc/info-service/get-trades.js#L37). Since both `getTrades` and `getOrderbook` are implementations for grpc methods, I think it's safe to assume that `limit` won't be undefined since the default protobuf value will be `'0'`, but let me know if you think it's valuable to keep the check for undefined.

I've updated `getOrderbook` to remove the check for `params.limit === undefined`.

Question: Would it be appropriate to also update `getTrades` to remove `params.limit === undefined` and the corresponding tests, or does that muddy the scope of this PR?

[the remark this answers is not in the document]
dannypaz: @bradleystachurski figured that we would do something like that XD. Feel free to remove it in this PR. I'll do a re-review and we can go from there.

broker-cli/proto/broker.proto
```diff
@@ -753,6 +753,8 @@ message MarketEvent {
 message GetOrderbookRequest {
   // Market to retrieve updates for, e.g. `'BTC/LTC'`.
   string market = 1;
+  // Number to limit results of either side of orderbook
+  int64 limit = 2;
```
treygriffith: consider naming this something like `limit_per_side` or similar. It may be a bit strange to have a request that you pass a limit to that returns 2x the number you specify.

broker-daemon/broker-rpc/orderbook-service/get-orderbook.js
```diff
@@ -25,6 +35,9 @@ async function getOrderbook ({ params, logger, orderbooks }, { GetOrderbookRespo
     const datetime = nano.toISOString(currentTime)
     const orders = await orderbook.all()
 
+    // limit is passed as an integer with default protobuf value of '0', so '0' implies limit was omitted
+    const limit = (params.limit === '0') ? DEFAULT_LIMIT : Big(params.limit)
```
treygriffith: this is kind of a bummer, but I guess there isn't a reason for someone to call this endpoint with a `limit` of 0. The other way to achieve this is with [`oneof`](https://developers.google.com/protocol-buffers/docs/proto3#oneof), although that may be overkill for this case.

broker-daemon/broker-rpc/orderbook-service/get-orderbook.js
```diff
@@ -36,6 +49,12 @@ async function getOrderbook ({ params, logger, orderbooks }, { GetOrderbookRespo
       }
     })
 
+    // Bids are sorted in descending order so the best bid (i.e. highest bid) is returned first
+    bids.sort((a, b) => parseFloat(b.price) - parseFloat(a.price)).splice(limit)
```
treygriffith: use `Big` instead of `parseFloat` here

martineehrlich: So I think we ideally want to the limit the number of records we get from the orderbook instead of getting all the records from the orderbook and then just returning a certain amount to the CLI. 

We have a method called `streamOrdersAtPriceOrBetter` which is defined on our PriceIndex. I think what you will want to do is something similar to what we have done in `orderbook.getBestOrders`, but instead of limited by depth, you will be limited by number of records. So your code here will look something like:

const bids = `orderbook.getOrders(bid, limit)`
const asks = `orderbook.getOrders(ask, limit)`

We could actually remove the iteration above: https://github.com/sparkswap/broker/pull/381/files#diff-7eb805b598ec65392196d352232266ffR43 and if we don't want to limit the number of records, just not pass in a limit.

And then `orderbook.getOrders` would look similar to `orderbook.getBestOrders` but you would use the limit to finish the stream when you got to the provided limit.

martineehrlich:
This looks great, just a few small comments. We can chat about the limit thing. My intuition is that the default should be unlimited because the user shouldn't have to specify a really large number to get all the records. If we have an issue with the number of records being too large, we can consider other options like pagination etc.

broker-cli/proto/broker.proto
```diff
@@ -859,6 +861,10 @@ service OrderBookService {
   rpc WatchMarket (WatchMarketRequest) returns (stream WatchMarketResponse);
   /**
    * GetOrderbook returns price and amount information about the bids and asks on the orderbook in the current state.
+   *
+   * When GetOrderbookRequest includes a limit parameter, the total number of bids or asks will not exceed
```
treygriffith: hmm, it's not the total number of bids and asks, its that the number of bids will not exceed the limit and the number of asks will not exceed the limit. Right?

broker-daemon/broker-rpc/orderbook-service/get-orderbook.spec.js
```diff
@@ -77,4 +85,41 @@ describe('getOrderbook', () => {
         timestamp: '1537487471431784972' }
     )
   })
+
+  it('limits the number of orders', async () => {
```
treygriffith: should add more `expect`s about calls to `getOrders` and the parameters you are expecting

broker-daemon/orderbook/index.js
```diff
@@ -164,6 +164,32 @@ class Orderbook {
     return trades
   }
 
+  /**
+   * Get orders in the orderbook for a given side up to a given limit. If no limit is provided, gets all orders
+   * @param {String} side Side of the orderbook to get orders for (i.e. `BID` or `ASK`)
+   * @param {String} limit int64 String of the the amount of orders to return.
+   * @return {Array<MarketEventOrder>} A promise that resolves MarketEventOrders for the limited records
+   */
+  getOrders ({ side, limit }) {
```
treygriffith: It seems like this is very similar to `all()`. Is there a way that we can remove `all()` and replace it with this new function? Where else is `all()` used?

broker-daemon/broker-rpc/orderbook-service/get-orderbook.js
```diff
@@ -1,10 +1,13 @@
 const nano = require('nano-seconds')
+const MarketEventOrder = require('../../models/market-event-order')
```
treygriffith: I think you should destructure `SIDES` from this since it's the only property you are using, e.g.

```
const { SIDES } = require('../../models/market-event-order')
```

also - is there a way to do this that doesn't require the rpc to reach into the models directly?

broker-daemon/broker-rpc/info-service/get-trades.js
```diff
@@ -34,7 +34,7 @@ async function getTrades ({ params, logger, orderbooks }, { GetTradesResponse })
     }
 
     // limit is passed as an integer so even if undefined it gets passed as '0', we can assume the limit should not be 0
-    const limit = (params.limit === '0' || params.limit === undefined) ? DEFAULT_LIMIT : Big(params.limit)
+    const limit = (params.limit === '0') ? DEFAULT_LIMIT : Big(params.limit)
```
dannypaz: @bradleystachurski lets not use `Big` because the way we usually access this value is through `.toString()` which would break the limit functionality of leveldb

PR #382, opened 2019-01-22
separate bitcoind/litecoind into own docker-compose files

martineehrlich:
## Description
We need to have an easy configuration setup if a user wants to use their own bitcoind/litecoind nodes. To do this, they need to supply the address of their nodes as env variables for:
```
BTC_ZMQPUBRAWBLOCK
BTC_ZMQPUBRAWTX
LTC_ZMQPUBRAWBLOCK
LTC_ZMQPUBRAWTX
```

Which have been defaulted in the env sample files.

We also do not want to build our own litecoind/bitcoind containers if the user is supplying their own, in which case bitcoind/litecoind containers should be added to docker-compose only if necessary. So I added  the COMPOSE_FILE env variable which specifies multiple docker-compose files that get composed together when calling docker-compose up. The user can remove the docker-compose.bitcoind.yml or docker-compose.litecoind.yml files from the path if they are supplying their own.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Link to Trello

### DIFF
--- a/.env-testnet-sample
+++ b/.env-testnet-sample
@@ -51,11 +51,12 @@ BTC_RPC_PASS=sparkswapbtc
 LTC_RPC_USER=sparkswapltc
 LTC_RPC_PASS=sparkswapltc
 
-# Location of nodes
+# Block Broadcast Addresses
 BTC_ZMQPUBRAWBLOCK=tcp://bitcoind:28333
-BTC_ZMQPUBRAWTX=tcp://bitcoind:28334
-
 LTC_ZMQPUBRAWBLOCK=tcp://litecoind:28333
+
+# Transaction Broadcast Addresses
+BTC_ZMQPUBRAWTX=tcp://bitcoind:28334
 LTC_ZMQPUBRAWTX=tcp://litecoind:28334
 
 # External IP Addresses
@@ -72,7 +73,7 @@ EXTERNAL_LTC_ADDRESS=sample.ip.address
 
 # Docker compose files
 # docker-compose by default will look for the docker-compose.yml file and apply the docker-compose.override.yml file if
-# it exists by default. If we specify the COMPOSE_FILE variable, all compose files joined by the COMPOSE_PATH_SEPARATOR will
+# it exists. If we specify the COMPOSE_FILE variable, all compose files joined by the COMPOSE_PATH_SEPARATOR will
 # be applied and used
 #
 COMPOSE_PATH_SEPARATOR=:
--- a/.env-testnet-sample
+++ b/.env-testnet-sample
@@ -36,8 +36,8 @@ RELAYER_CERT_HOST=relayer.testnet.sparkswap.com:8080
 # RPC_HOST. You will be required to also set different RPC_USER/RPC_PASS values
 #
 # Daemon RPC Hosts
-BTC_RPC_HOST=bitcoind
-LTC_RPC_HOST=litecoind
+# BTC_RPC_HOST=your.btc.rpc.host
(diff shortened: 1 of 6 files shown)


.env-testnet-sample
```diff
@@ -62,3 +69,11 @@ LTC_RPC_PASS=sparkswapltc
 #
 EXTERNAL_BTC_ADDRESS=sample.ip.address
 EXTERNAL_LTC_ADDRESS=sample.ip.address
+
+# Docker compose files
+# docker-compose by default will look for the docker-compose.yml file and apply the docker-compose.override.yml file if
```
dannypaz: weird sentence cause we say `by default` twice.

docker-compose.litecoind.yml
```diff
@@ -0,0 +1,45 @@
...
+      - DATA_DIR=/data
+      - RPC_LISTEN=0.0.0.0
+    networks:
+      - broker
+    logging:
+      options:
+        max-size: 50m
+
+volumes:
+  shared:
```
treygriffith: Is it ok to declare this volume twice like this? Is it the same volume that's declared in the other file?

.env-testnet-sample
```diff
@@ -62,3 +69,11 @@ LTC_RPC_PASS=sparkswapltc
...
+# it exists by default. If we specify the COMPOSE_FILE variable, all compose files joined by the COMPOSE_PATH_SEPARATOR will
+# be applied and used
+#
+COMPOSE_PATH_SEPARATOR=:
```
treygriffith: Is there a reason that `:` isn't the path separator on windows? Will this work on windows?

treygriffith:
@martineehrlich super happy that you picked up this ticket 👏

dannypaz:
@martineehrlich My changes look good. @treygriffith will need to sign off

.env-testnet-sample
martineehrlich: https://github.com/docker/compose/pull/4588, COMPOSE_PATH_SEPARATOR was added so that it uses this as the separator no matter what.

docker-compose.litecoind.yml
martineehrlich: Yes, its fine. It just merges all of the volumes together. We also could remove it in all but docker-compose.yml, but I think it is confusing to have volumes defined above and then not have them below. It also makes it so you could never independently run the specific docker-compose file. 

https://docs.docker.com/compose/extends/#adding-and-overriding-configuration

.env-testnet-sample
```diff
@@ -62,3 +70,11 @@ LTC_RPC_PASS=sparkswapltc
...
 EXTERNAL_BTC_ADDRESS=sample.ip.address
 EXTERNAL_LTC_ADDRESS=sample.ip.address
+
+# Docker compose files
+# docker-compose by default will look for the docker-compose.yml file and apply the docker-compose.override.yml file if
+# it exists. If we specify the COMPOSE_FILE variable, all compose files joined by the COMPOSE_PATH_SEPARATOR will
+# be applied and used
+#
+COMPOSE_PATH_SEPARATOR=:
+COMPOSE_FILE=docker-compose.yml:docker-compose.bitcoind.yml:docker-compose.litecoind.yml:docker-compose.override.yml
```
treygriffith: Is it possible for us to use other variables to build up this variable? e.g. have a USE_BUNDLED_BITCOIND variable that if set to true, includes docker-compose.bitcoind.yml in the path, but otherwise doesn't.

PR #383, opened 2019-01-23
FEATURE - Broker Docker Image

dannypaz:
## Description
This PR modifies our current broker docker image to allow for hosting on hub.docker.com. The following modifications were made:

1. Removed certificate generation in Dockerfile (this is now passed as an env when the image is used)
2. Removed TLS cert copying from Dockerfile (this is now specified as an env)
3. Reduced image size by > 60% (using alpine for both build and final images)
    - Previous build image was 1.2gb+, now reduced to 480mb
    - Previous final image was 900mb, now reduced to 264mb
4. Added option flags for `npm run build`
    - `-c, --no-cli`
    - `-d, --no-docker`
    - `-e=*, --external-address=*`
    - `-i, --no-identity` - controls generation of ID pub/priv keys for broker daemon
    - `-n, --no-certs` - controls TLS cert generation for broker daemon/broker-cli
5. Removed `npm install` on `npm run build` as it takes a considerable amount of time to index for NPM when the call is basically redundant (1. we are using docker 2. you only need deps for tests)
6. Simplified our build scripts by removing `prebuild.sh` and `build-images` in favor of build options
7. Persistence checks of TLS and ID certs/keys inside of `build.sh`

#### User Requirements

1. The user will need to generate their own ID and TLS certs (on their own machine w/ `npm run build`)
    - We already make the user do this for a normal setup
    - It is not possible to generate unique certs in a prebuilt docker image (IMPORTANT)
    - We want the user to have IDs and TLS certs persisted through the setup, it was then redundant to have the docker image generate these certs
    - The user can specify the cert path through ENV variables (if copied into the container)
    - The user can specify the cert directory through bind mounting when running `docker run`
    - The user can specify the cert directory through bind mounting BY DEFAULT through `docker-compose.yml`
2. The user should be able to use the broker-cli remotely
    - This is covered by the cert generation and was a regression we were trying to avoid
3. The setup should be as simple as possible
    - Certs are required for the setup
    - Defaults exist for everything else

There are probably system compatibility issues that will come up regarding `openssl`, but this is something that we can get to when the time comes

#### Migrating from an old setup

In this PR, we have updated environment and docker-compose files which require us to rebuild the docker container after we pull in these changes. Run the following commands below:

```
mv .env .env.bak
mv docker-compose.override.yml docker-compose.override.yml.bak
cp .env-regtest-sample .env
cp docker-compose.regtest.sample.yml docker-compose.override.yml
npm run build
docker-compose up -d
```

Once you have confirmed that the setup is functioning, you can remove the backup override/env files. You do not need to backup the env/override files if you have not made any custom changes

## Related PRs

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Link to Trello


dannypaz:
[Create Standalone Broker Docker Image](https://trello.com/c/rL15yr8m/1072-create-standalone-broker-docker-image)

scripts/build.sh
```diff
@@ -3,110 +3,187 @@
...
+
+    ;;
+    *)
+            # unknown option
+    ;;
+esac
+done
+
+if [ "$EXTERNAL_ADDRESS" == "" ]; then
+  echo "Please provide your public IP address"
```
treygriffith: or hostname

scripts/build.sh
```diff
@@ -3,110 +3,187 @@
...
 
 echo ""
 echo "It's time to BUILD! All resistance is futile."
 echo ""
 
-echo "Reinstalling dependencies"
-npm install
+if [ "$PRODUCTION" == "true" ]; then
+  echo "Reinstalling PRODUCTION dependencies"
+  npm install --quiet --production
```
treygriffith: we don't install anything outside of production? I would think that we install but without the production flag

scripts/build.sh
```diff
@@ -3,110 +3,187 @@
...
+fi
+
+if [ -f docker-compose.override.yml ]; then
   # Let the user know that an override file exists which may mean that the user
   # will have settings they do not expect
-  echo "A 'docker-compose.override.yml' file exists, but you are not using a supported"
-  echo "environment: local or regtest"
   echo ""
+  echo "WARNING: A 'docker-compose.override.yml' file exists"
   echo "WARNING: This may add unwanted settings to the broker that could effect how your daemon runs."
```
treygriffith: "affect", not "effect"

scripts/build.sh
```diff
@@ -3,110 +3,187 @@
...
 fi
 
-echo "Copying certs"
-
 echo "Making local ~/.sparkswap certs directory"
 DIRECTORY=~/.sparkswap
 mkdir -p $DIRECTORY/certs
 
 echo "Copying certs to local certs directory"
+# Not sure what to do here because the certs is gonna fuck everything up
```
treygriffith: probably not a comment we want in the codebase

treygriffith:
With this PR, what are the two possible user flows (for using the pre-built image or building yourself)?

I don't think I fully understand how a user uses these changes.

docker/sparkswapd/Dockerfile
```diff
@@ -1,79 +1,47 @@
...
 
+# Copy node_modules and rebuild
+COPY --from=builder /home/app/node_modules ./node_modules
+
 WORKDIR /home/app
 
-RUN npm install --quiet
-RUN npm run broker-proto
+RUN npm rebuild -q
+RUN npm install pm2 -g
```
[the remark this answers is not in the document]
dannypaz: @treygriffith this is a good point. we should move PM2 to dependencies in `package.json` instead of dev dependencies (we use --production for npm above)

scripts/build.sh
dannypaz: @treygriffith the only time you would ever need to install dependencies is for tests. The user installs the broker-cli separately (and that installs the correct deps), so really... the only time we need `npm install` is for dev machines

dannypaz: @treygriffith Maybe this isnt the correct way to go about it but i've omitted the installation line because it is redundant (you'll npm install locally, then npm install inside of the image)

scripts/build.sh
dannypaz: woops, that is my bad. I will rebase

docker/sparkswapd/Dockerfile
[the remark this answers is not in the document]
dannypaz: @treygriffith this was actually a huge bug. I wasn't copying `node_modules` to the correct folder in the docker image which is why `pm2` was not present (and required me to globally install pm2). Thanks for the catch

scripts/build.sh
```diff
@@ -3,110 +3,178 @@
 ################################################
 # Build script for sparkswapd
 #
-# Params:
-# - EXTERNAL_ADDRESS (optional, for hosted daemons)
+# Options:
```
treygriffith: I think you need to update the `env-setup` and `install.sh` scripts with these new options and make sure everything still works together

treygriffith:
It might be worthwhile to do a clean build before and after this change to get metrics on:
- build time saved
- build size reduced

Of course the big savings are from being able to have a prebuilt image, but I think it would be interesting.

docker-compose.yml
```diff
@@ -45,7 +47,7 @@ services:
     networks:
       - broker
     working_dir: '/home/app'
-    entrypoint: bash -c 'npm run start-sparkswapd'
+    entrypoint: bash -c 'npm run start-sparkswapd-watch'
```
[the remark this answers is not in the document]
dannypaz: @treygriffith good catch. This can actually go in the `regtest` sample docker-compose file. We need this line so that we can watch the file system for changes (hot reloading for pm2)

docker/sparkswapd/Dockerfile
```diff
@@ -1,79 +1,46 @@
...
-#############################################
-
-# Copy local broker files to container
-COPY . /home/app
+WORKDIR /home/app
 
-# Download relayer proto and add it to container
-ARG RELAYER_PROTO_BRANCH
-ARG RELAYER_PROTO_COMMIT_SHA
+RUN git clone -b ${BROKER_VERSION} https://github.com/sparkswap/broker.git .
```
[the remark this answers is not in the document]
dannypaz: @treygriffith Not exactly. If you use the `regtest.sample` yml file in your setup, then our bind mounts override the `git clone`.

The only drawback to this approach is that the container _could_ have super stale files if we don't increment the version periodically (which invalidates the cache).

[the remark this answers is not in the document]
dannypaz: The plus side to this change is that we can publish this docker file directly to hub.docker.com (which is why the change was made)

dannypaz:
@treygriffith got you some metrics on the builds.

Tests were performed with `time npm run build`.
Before beginning the tests I stopped all containers and cleared all images with `docker rm $(docker ps -qa)` and `docker image rm $(docker image ls -q)`


**Old Setup:**
Time taken for nom run build: `3:38`

```
REPOSITORY             TAG                 IMAGE ID            CREATED              SIZE
sparkswap_sparkswapd   latest              69c1a5beffca        55 seconds ago       963MB
<none>                 <none>              3dce821221e8        About a minute ago   969MB
node                   8.11                8198006b2b57        4 months ago         673MB
```

**New Setup:**
Time taken for npm run build: `2:12`

```
REPOSITORY             TAG                 IMAGE ID            CREATED             SIZE
sparkswap_sparkswapd   latest              a713ce30b1b8        5 minutes ago       236MB
<none>                 <none>              41ab4b4d6d64        5 minutes ago       494MB
node                   8.11-alpine         8adf3c3eb26c        4 months ago        68.1MB
```


> Sparkswapd image is 75% smaller (24.5% of the previous size)
> Broker images total size is 70% smaller (30% of the previous size)

PR #386, opened 2019-01-30
Register public key with relayer authorization service

martineehrlich:
## Description
The Broker needs to expose an RPC to register it's public key with the Relayer by calling the Relayer's Register rpc.

This PR adds the register CLI command that takes in a public key and returns the entity id of the entity created by the relayer authorization service.
The broker register RPC calls the relayers register RPC.

## Related PRs
https://github.com/sparkswap/relayer/pull/223

## Todos
- [x] Tests
- [x] Documentation
- [x] Link to Trello

### DIFF
--- a/broker-cli/commands/register.js
+++ b/broker-cli/commands/register.js
@@ -4,7 +4,7 @@ const { RPC_ADDRESS_HELP_STRING } = require('../utils/strings')
 
 /**
  *
- * Register publicKey with the Relayer's authorization service
+ * Register publicKey with the Relayer
  *
  * ex: `sparkswap register`
  *
@@ -20,10 +20,9 @@ async function register (args, opts, logger) {
   try {
     const client = await new BrokerDaemonClient(rpcAddress)
 
-    // Currently we don't do anything with this entityId but we will need it in the future
-    const { entityId } = await client.adminService.register({})
+    await client.adminService.register({})
 
-    logger.info('Successfully registered', { entityId })
+    logger.info('Successfully registered public key with relayer')
   } catch (e) {
     logger.error(handleError(e))
   }
@@ -31,7 +30,7 @@ async function register (args, opts, logger) {
 
 module.exports = (program) => {
   program
-    .command('register', 'Registers the public key with the authorization service')
+    .command('register', 'Registers the public key with the relayer')
     .option('--rpc-address [rpc-address]', RPC_ADDRESS_HELP_STRING, validations.isHost)
     .action(register)
 }
--- a/broker-cli/commands/register.js
+++ b/broker-cli/commands/register.js
@@ -3,11 +3,10 @@ const { validations, handleError } = require('../utils')
 const { RPC_ADDRESS_HELP_STRING } = require('../utils/strings')
 
 /**
- * sparkswap register
  *
  * Register publicKey with the Relayer's authorization service
  *
- * ex: `sparkswap id`
+ * ex: `sparkswap register`
  *
  * @param {Object} args
  * @param {Object} opts
--- a/broker-cli/commands/register.js
+++ b/broker-cli/commands/register.js
@@ -3,29 +3,28 @@ const { validations, handleError } = require('../utils')
 const { RPC_ADDRESS_HELP_STRING } = require('../utils/strings')
 
 /**
- * sparkswap id
+ * sparkswap register
  *
- * Get the BrokerDaemon's Identity
+ * Register publicKey with the Relayer's authorization service
  *
  * ex: `sparkswap id`
  *
  * @param {Object} args
- * @param {String} args.publicKey
  * @param {Object} opts
  * @param {String} [rpcAddress=null] broker rpc address
  * @param {Logger} logger
  */
 
 async function register (args, opts, logger) {
-  const { publicKey } = args
   const { rpcAddress = null } = opts
 
   try {
     const client = await new BrokerDaemonClient(rpcAddress)
 
-    const { entityId } = await client.adminService.register({publicKey})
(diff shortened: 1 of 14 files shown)


martineehrlich:
[[Auth] Add register broker rpc](https://trello.com/c/5PQcXWNe/1081-auth-add-register-broker-rpc)

broker-cli/commands/register.js
```diff
@@ -0,0 +1,40 @@
...
+ *
+ * @param {Object} args
+ * @param {String} args.publicKey
+ * @param {Object} opts
+ * @param {String} [rpcAddress=null] broker rpc address
+ * @param {Logger} logger
+ */
+
+async function register (args, opts, logger) {
+  const { publicKey } = args
```
dannypaz: Im thinking that there is probably a way that `publicKey` can become an option instead of being an argument for the register command... because we already know where the path to this key is.

`sparkswap register` would grab the key from the path that was specified in the configuration file?

treygriffith: Yeah I agree with this, the `publicKey` should probably be a path, and have a default to somewhere in the `.sparkswap` directory

treygriffith: But to be clear, I think that if we're loading from a path, that should be done within the daemon, not from the CLI. (I think, someone check my thinking on that)

broker-daemon/broker-rpc/admin-service/register.js
```diff
@@ -0,0 +1,19 @@
...
+ * @param {RelayerClient} request.relayer - grpc Client for interacting with the Relayer
+ * @param {Object} request.logger
+ * @param {Object} responses
+ * @param {function} responses.RegisterResponse - constructor for RegisterResponse messages
+ * @return {RegisterResponse}
+ */
+async function register ({ params, relayer, logger }, { RegisterResponse }) {
+  const { publicKey } = params
+
+  const { entityId } = await relayer.adminService.register({publicKey})
```
treygriffith: I think we should probably persist this `entityId` id somewhere. Like in the `.sparkswap` directory

broker-cli/commands/register.js
```diff
@@ -0,0 +1,40 @@
+const BrokerDaemonClient = require('../broker-daemon-client')
+const { validations, handleError } = require('../utils')
+const { RPC_ADDRESS_HELP_STRING } = require('../utils/strings')
+
+/**
+ * sparkswap id
+ *
+ * Get the BrokerDaemon's Identity
+ *
+ * ex: `sparkswap id`
```
treygriffith: these docs are for another command

broker-cli/commands/register.js
```diff
@@ -0,0 +1,40 @@
...
+async function register (args, opts, logger) {
+  const { publicKey } = args
+  const { rpcAddress = null } = opts
+
+  try {
+    const client = await new BrokerDaemonClient(rpcAddress)
+
+    const { entityId } = await client.adminService.register({publicKey})
+
+    logger.info(entityId)
```
treygriffith: Perhaps some more information here, like "successfully registered"?

broker-cli/utils/validations.js
```diff
@@ -6,6 +6,13 @@ const { Big } = require('./big')
  */
 const MAX_VALUE = '9223372036854775807'
 
+/**
+ * RegExp for matching Base64 strings, which public keys are (once stripped of newlines and banners)
+ * @constant
+ * @type {RegExp}
+ */
+const BASE64_MATCH = /^(?:[A-Za-z0-9+/]{4})*(?:[A-Za-z0-9+/]{2}==|[A-Za-z0-9+/]{3}=)?$/
```
dannypaz: just for reference, looks like https://stackoverflow.com/a/475217/1670474 is where the regex came from

broker-cli/proto/broker.proto
```diff
@@ -149,6 +149,11 @@ message GetIdentityResponse {
   string public_key = 1;
 }
 
+message RegisterResponse {
+  // The entity_id of the entity registered with the authorization service
```
treygriffith: The Broker probably shouldn't have a notion of the authorization service, I would just say that this is "the Identifier assigned by the Relayer for the Entity associated with the provided Public Key".

broker-daemon/broker-rpc/admin-service/register.js
```diff
@@ -0,0 +1,19 @@
...
+ * @param {RelayerClient} request.relayer - grpc Client for interacting with the Relayer
+ * @param {Object} request.logger
+ * @param {Object} responses
+ * @param {function} responses.RegisterResponse - constructor for RegisterResponse messages
+ * @return {RegisterResponse}
+ */
+async function register ({ relayer, logger }, { RegisterResponse }) {
+  const publicKey = relayer.identity.pubKeyBase64
+
+  const { entityId } = await relayer.adminService.register({publicKey})
```
treygriffith: [nit] can you put a little breathing room around `publicKey`

broker-daemon/broker-rpc/admin-service/register.js
```diff
@@ -0,0 +1,20 @@
...
+ * @param {Object} responses
+ * @param {function} responses.RegisterResponse - constructor for RegisterResponse messages
+ * @return {RegisterResponse}
+ */
+async function register ({ relayer, logger }, { RegisterResponse }) {
+  const publicKey = relayer.identity.pubKeyBase64
+
+  // Currently we don't do anything with this entityId but we will need it in the future
+  const { entityId } = await relayer.adminService.register({ publicKey })
+
```
treygriffith: might be worthwhile to add a log here saying "successfully registered with relayer and include the entity id

PR #389, opened 2019-01-31
Add authorization into metadata

treygriffith:
## Description
This change moves our authorization data into the metadata of a request rather than its body.

## Related PRs

## Todos
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/broker-daemon/relayer/identity.js
+++ b/broker-daemon/relayer/identity.js
@@ -53,33 +53,21 @@ class Identity {
   }
 
   /**
-   * @typedef {Object} Authorization
-   * @property {String} timestamp Int64 string of the current unix timestamp in seconds
-   * @property {String} nonce base64 string of 32 random bytes
-   * @property {String} signature base64 string of the signature that authorizes it
-   */
-
-  /**
    * Authorize a gRPC Request by adding metadata to it.
    *
    * Specifically, the signature is of the following payload:
-   *  - a string name of the request [ what format? ]
    *  - timestamp of the request, in seconds. A timestamp within +/- 60 seconds of the Relayer's time should be accepted.
    *  - A random nonce of 32 bytes, represented in base64. This nonce should not be repeated, but the Relayer may not reject duplicate nonces older than 24 hours.
-   *  - Contents of the request, JSON stringified
    *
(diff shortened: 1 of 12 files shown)


broker-daemon/relayer/identity.js
```diff
@@ -77,30 +60,33 @@ class Identity {
...
    *
    * The request can then be validated by the Relayer as being genuine from the owner of the public key.
-   * @param  {String} id - id to authorize a request for
-   * @return {Authorization}
+   * @param  {string} action action to authorize the request for
+   * @param  {object} params Parameters of the request to authorize
+   * @return {grpc.Metadata} Metadata object with necessary items to verify it
    */
-  authorize (id) {
+  authorize (action, params) {
```
martineehrlich: Why is the action necessary? Aren't we already going to be hitting the endpoint for the action? It just feels messy to have the code littered with action strings with no good place to centralize them, especially because the actions seem redundant (unless I'm missing something)

dannypaz: @treygriffith I think this brings up a fair point. I was actually debating the same thing on my PR in the relayer (although for the relayer, it is a little bit more sane).

We wanted to use `action` as a deterrent of replaying the request, however the signature itself (with timestamp nonce etc) should be enough data that the `action` isn't required.

On the relayer side, we'll need to worry about the action, but we already have this information when the request comes so theres no real benefit to having it on the broker if we dont need it

dannypaz:
@treygriffith other than the comment about, this is a good change. We should definitely talk about what we want to do for the `signature`

broker-daemon/relayer/identity.js
```diff
@@ -70,37 +53,28 @@ class Identity {
...
     const signature = this.sign(payload)
+    const pubKey = this.pubKeyBase64
 
-    return {
-      timestamp,
-      nonce,
-      signature
-    }
+    const metadata = new Metadata()
+    metadata.set('authorization', `${pubKey},${timestamp},${nonce},${signature}`)
```
dannypaz: did you not want to set pubkey/timestamp/nonce as separate fields for metadata? Then we wouldn't have to parse this information (via string) on the server

dannypaz:
@treygriffith I also tested this PR. LGTM

PR #390, opened 2019-01-31
Fix - Disable auth not working

dannypaz:
## Description
I omitted the `[disable-auth]` parameter option from the `--disable-auth` option of `sparkswapd` which prevents the defaults in `config.json` from working correctly.

**What is currently happening:**
If you omit the `--disable-auth` flag when starting `sparkswapd` and `config.json` specifies `disableAuth: true`, the parameter will still be false because the flag is not present on sparkswapd (the default will be ignored)

**What should happen:**
If you set `disableAuth: true` in config.json, it should disable auth

## Related PRs

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Link to Trello

### DIFF
--- a/broker-cli/proto/broker.proto
+++ b/broker-cli/proto/broker.proto
@@ -787,6 +787,8 @@ message MarketEvent {
 message GetOrderbookRequest {
   // Market to retrieve updates for, e.g. `'BTC/LTC'`.
   string market = 1;
+  // Number to limit results of either side of orderbook
+  int64 limit_per_side = 2;
 }
 
 message Bid {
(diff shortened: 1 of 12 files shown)


dannypaz:
[disable-auth option not working correctly](https://trello.com/c/lSBymCk4/1113-disable-auth-option-not-working-correctly)

treygriffith:
Per our conversation: Caporal doesn't allow implicit booleans (e.g. `--disable-auth` vs `--disable-auth [disable-auth]`) to have a default value, which makes sense.

What that really means for us is that our use of caporal defaults as a way to inject configuration doesn't really work the way we want it to.

For now, I think this is fine, but we should re-examine at a future point if there is a better way to apply the configuration file and merge it with command-line arguments.

PR #391, opened 2019-01-31
Persist certs by having them in .sparkswap directory

martineehrlich:
## Description
We want the rpc and identity certs to be in a persistent location (so we don't regenerate them every time we build)

This PR:
- Puts the generated certs in the `.sparkswap/certs` directory if they don't exist
- Binds the /secure folder in the container to the cert path specified in the .env file (default is `.sparkswap/certs`)

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Link to Trello

### DIFF
--- a/.env-mainnet-sample
+++ b/.env-mainnet-sample
@@ -68,5 +68,5 @@ EXTERNAL_LTC_ADDRESS=sample.ip.address
 COMPOSE_PATH_SEPARATOR=:
 COMPOSE_FILE=docker-compose.yml:docker-compose.bitcoind.yml:docker-compose.litecoind.yml:docker-compose.override.yml
 
-# Path to rpc and identity certs
-PATH_TO_CERTS=~/.sparkswap/certs
+# Path to rpc certs and identity keys
+SECURE_PATH=~/.sparkswap/secure
--- a/.env-mainnet-sample
+++ b/.env-mainnet-sample
@@ -60,6 +60,9 @@ LTC_RPC_PASS=sparkswapltc
 EXTERNAL_BTC_ADDRESS=sample.ip.address
 EXTERNAL_LTC_ADDRESS=sample.ip.address
 
+# Path to rpc certs and identity keys
+SECURE_PATH=~/.sparkswap/secure
+
 # Docker compose files
 # docker-compose by default will look for the docker-compose.yml file and apply the docker-compose.override.yml file if
 # it exists. If we specify the COMPOSE_FILE variable, all compose files joined by the COMPOSE_PATH_SEPARATOR will
@@ -67,6 +70,3 @@ EXTERNAL_LTC_ADDRESS=sample.ip.address
 #
 COMPOSE_PATH_SEPARATOR=:
 COMPOSE_FILE=docker-compose.yml:docker-compose.bitcoind.yml:docker-compose.litecoind.yml:docker-compose.override.yml
-
-# Path to rpc certs and identity keys
-SECURE_PATH=~/.sparkswap/secure
--- a/.env-mainnet-sample
+++ b/.env-mainnet-sample
@@ -67,3 +67,6 @@ EXTERNAL_LTC_ADDRESS=sample.ip.address
 #
 COMPOSE_PATH_SEPARATOR=:
 COMPOSE_FILE=docker-compose.yml:docker-compose.bitcoind.yml:docker-compose.litecoind.yml:docker-compose.override.yml
+
+# Path to rpc and identity certs
+PATH_TO_CERTS=~/.sparkswap/certs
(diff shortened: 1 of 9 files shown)


scripts/build.sh
```diff
@@ -97,9 +97,22 @@ rm -rf ./proto/.git
...
+  mkdir -p $DIRECTORY
+fi
+
+if [[ ! -d "$DIRECTORY/certs" ]]; then
+  echo "Creating directory $DIRECTORY/certs"
+
+  mkdir -p $DIRECTORY/certs
+fi
+
+KEY_PATH=~/.sparkswap/certs/broker-rpc-tls.key
```
treygriffith: why aren't we using `DIRECTORY` here?

scripts/build.sh
```diff
@@ -97,9 +97,22 @@ rm -rf ./proto/.git
 # Primary use is TLS between Broker-CLI and Broker Daemon
 #
 #############################################
-KEY_PATH="./certs/broker-rpc-tls.key"
-CERT_PATH="./certs/broker-rpc-tls.cert"
-CSR_PATH="./certs/broker-rpc-csr.csr"
+
+DIRECTORY=~/.sparkswap
```
treygriffith: it feels like this should be configurable, perhaps via an environment variable or command-line argument.

docker/sparkswapd/Dockerfile
```diff
@@ -15,7 +15,7 @@ WORKDIR /home/app
 RUN git clone -b ${BROKER_VERSION} https://github.com/sparkswap/broker.git .
 
 RUN npm install --quiet --production
-RUN npm run build -- --no-docker
+RUN npm run build -- --no-docker --no-identity --no-certs
```
treygriffith: [nit] might be worth adding a comment above as to why we pass these flags when building the image

.env-mainnet-sample
```diff
@@ -67,3 +67,6 @@ EXTERNAL_LTC_ADDRESS=sample.ip.address
 #
 COMPOSE_PATH_SEPARATOR=:
 COMPOSE_FILE=docker-compose.yml:docker-compose.bitcoind.yml:docker-compose.litecoind.yml:docker-compose.override.yml
+
+# Path to rpc and identity certs
```
treygriffith: The identity is not a cert, just a private and public key. Maybe we should rename the directory that they are in as well?

scripts/build.sh
martineehrlich: We should 😨

.env-mainnet-sample
martineehrlich: to keys?

scripts/build.sh
dannypaz: Just to add documentation to this. Since we do not allow the `~/.sparkswap` directory to be configurable for the CLI, we've instead added the task `Make ~/.sparkswap application directory configurable` to an upcoming roadmap

docker-compose.yml
```diff
@@ -25,8 +25,9 @@ services:
       - 'sparkswapd:/data'
       # This is populated externally w/ an engine
       - 'shared:/shared'
-      # Persistent certs for broker
-      - './certs:/secure'
+      # Persistent certs/keys for broker
+      - ${SECURE_PATH}:/secure
```
dannypaz: @martineehrlich just curious, did you get any more info on `""` vs `''` vs no quotes?

.env-mainnet-sample
```diff
@@ -67,3 +67,6 @@ EXTERNAL_LTC_ADDRESS=sample.ip.address
 #
 COMPOSE_PATH_SEPARATOR=:
 COMPOSE_FILE=docker-compose.yml:docker-compose.bitcoind.yml:docker-compose.litecoind.yml:docker-compose.override.yml
+
+# Path to rpc certs and identity keys
+SECURE_PATH=~/.sparkswap/secure
```
dannypaz: [nit] suchhh a nit, but would rather all other information above the `COMPOSE_FILE` lines (having the COMPOSE lines be the last lines of the file)

treygriffith: agree

scripts/build.sh
treygriffith: [nit] maybe put this at the top of the file and call it something more specific like `SPARKSWAP_DIRECTORY` or similar?

scripts/build.sh
```diff
@@ -97,9 +97,22 @@ rm -rf ./proto/.git
...
+if [ ! -d "$DIRECTORY" ]; then
```
treygriffith: `mkdir -p` won't create the directory if it doesn't exist (and I think it is recursive), so why not just have this and the below like:

```
echo "Creating directories $DIRECTORY and $DIRECTORY/secure"
mkdir -p $DIRECTORY/secure
```

PR #392, opened 2019-01-31
Fix - GRPC Gateway

dannypaz:
## Description
This PR fixes a breaking change that was introduced in https://github.com/sparkswap/broker/pull/354 where we incorrectly pass in `rpcHttpProxyAddress` into `grpc-gateway` instead of the `rpcAddress`. The passing of an incorrect address broke the http proxy completely.

**Changes**
1. format some hard to read places in `grpc-gateway.js`
2. pass the correct `rpcAddress` to grpc-gateway
3. update http server address handling (changing 0.0.0.0 to localhost for https AND http)
4. add 404 handler for https routes

## Related PRs

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Link to Trello

### DIFF
--- a/broker-daemon/config.json
+++ b/broker-daemon/config.json
@@ -5,7 +5,7 @@
   "idPrivKeyPath": "/secure/broker-identity.private.pem",
   "markets": "BTC/LTC",
   "//": "WARNING: Only disable broker SSL during development",
-  "disableAuth": false,
+  "disableAuth": true,
   "supportedEngineTypes": ["LND"],
   "relayer": {
     "rpcHost": "host.docker.internal:28492",
--- a/broker-daemon/config.json
+++ b/broker-daemon/config.json
@@ -5,7 +5,7 @@
   "idPrivKeyPath": "/secure/broker-identity.private.pem",
   "markets": "BTC/LTC",
   "//": "WARNING: Only disable broker SSL during development",
-  "disableAuth": true,
+  "disableAuth": false,
   "supportedEngineTypes": ["LND"],
   "relayer": {
     "rpcHost": "host.docker.internal:28492",
--- a/broker-daemon/utils/create-http-server.js
+++ b/broker-daemon/utils/create-http-server.js
@@ -40,6 +40,14 @@ function createHttpServer (protoPath, rpcAddress, { disableAuth = false, enableC
 
   if (disableAuth) {
     app.use('/', grpcGateway([`/${protoPath}`], rpcAddress))
+
+    // Handle 404s correctly for the server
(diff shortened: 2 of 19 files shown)


broker-daemon/utils/create-http-server.js
```diff
@@ -30,24 +30,30 @@ function createHttpServer (protoPath, rpcAddress, { disableAuth = false, enableC
     app.use(corsMiddleware())
   }
 
+  // If the RPC address we use for daemon is set to a default route (0.0.0.0)
```
dannypaz: I moved these lines to the top of the file, because we still want to mutate this address whether disableAuth is true or not

broker-daemon/utils/create-http-server.js
```diff
@@ -30,24 +30,30 @@ function createHttpServer (protoPath, rpcAddress, { disableAuth = false, enableC
...
-    // to local host since 0.0.0.0 would be an invalid address w/ the current cert
-    // setup
-    if (rpcAddress.includes('0.0.0.0')) {
-      rpcAddress = rpcAddress.replace('0.0.0.0', 'localhost')
-    }
-
     app.use('/', grpcGateway([`/${protoPath}`], rpcAddress, channelCredentials))
 
-    logger.debug(`Securing http RPC connections with TLS: key: ${privKeyPath}, cert: ${pubKeyPath}`)
+    // Handle 404s correctly for the server
```
dannypaz: This handles 404s for `https` which was missing (we only handled 404s for http)

dannypaz:
[grpc gateway not accepting requests](https://trello.com/c/LJHUxGSg/1115-grpc-gateway-not-accepting-requests)

broker-daemon/utils/create-http-server.js
treygriffith: should we have a single handler before we do the http/https divide?

dannypaz: @treygriffith theres probably a way I can refactor this to make it nicer, but the `404` handling has to happen as the last action after grpc gateway, so it makes the code a little weird

broker-daemon/config.json
```diff
@@ -5,7 +5,7 @@
   "idPrivKeyPath": "/secure/broker-identity.private.pem",
   "markets": "BTC/LTC",
   "//": "WARNING: Only disable broker SSL during development",
-  "disableAuth": false,
+  "disableAuth": true,
```
treygriffith: This seems like it was left in by mistake

dannypaz:
@treygriffith good catch. reverted the change. thanks for reviewing

PR #407, opened 2019-02-12
move info service commands into orderbook service

martineehrlich:
## Description
The cli commands in the InfoService are about the market and can go under the OrderbookService instead. Along with moving the InfoService -> OrderbookService, I also renamed the info command to be `sparkswap market {insertcall}`

## Todos
- [x] Tests
- [x] Documentation
- [x] Link to Trello


martineehrlich:
[Move InfoService calls to the OrderbookService](https://trello.com/c/NU7RxKLx/1106-move-infoservice-calls-to-the-orderbookservice)

broker-cli/commands/info/index.js
```diff
@@ -3,11 +3,11 @@ const { RPC_ADDRESS_HELP_STRING, MARKET_NAME_HELP_STRING } = require('../../util
 
 /**
  * Info
```
dannypaz: would rename this folder name to markets and fix the text in this comment

broker-daemon/relayer/relayer-client.js
```diff
@@ -55,7 +55,7 @@ class RelayerClient {
 
     this.makerService = caller(this.address, this.proto.MakerService, this.credentials)
     this.takerService = caller(this.address, this.proto.TakerService, this.credentials)
-    this.orderbookService = caller(this.address, this.proto.OrderBookService, this.credentials)
+    this.orderBookService = caller(this.address, this.proto.OrderBookService, this.credentials)
```
dannypaz: @martineehrlich @treygriffith we should make a note that we want to use `Order Book` (2 words) instead of `orderbook` (one). We are inconsistent in this text in a ton of places

broker-cli/commands/market/index.js
```diff
@@ -3,11 +3,11 @@ const { RPC_ADDRESS_HELP_STRING, MARKET_NAME_HELP_STRING } = require('../../util
 
 /**
  * Info
```
dannypaz: change this text to market

dannypaz:
@martineehrlich other than fixing the info comment, this looks good to me. I've tested this on my broker and the commands are +1.

dannypaz:
FWIW this will be a breaking change, so it will be worth it to put this into release notes. I've written a ticket for the next release so we can handle that.

PR #408, opened 2019-02-14
add function for cancelling all active block orders for a given market

martineehrlich:
## Description
When a user releases their channels, we should cancel all their orders on that market because the user will no longer have the funds to fulfill those orders.

## Todos
- [x] Tests
- [x] Documentation
- [x] Link to Trello

### DIFF
--- a/broker-daemon/block-order-worker/index.js
+++ b/broker-daemon/block-order-worker/index.js
@@ -359,6 +359,31 @@ class BlockOrderWorker extends EventEmitter {
   }
 
   /**
+   * Cancel all active orders for a given market
+   * @param  {string} market to cancel orders on
+   * @return {object} successfulCancellations and unsuccessfulCancellations as arrays of blockorderids
+   */
+  async cancelActiveOrders (market) {
+    this.logger.info('Cancelling all active orders for market', { market })
+
+    const blockOrders = await this.getBlockOrders(market)
+    const activeBlockOrders = blockOrders.filter(blockOrder => blockOrder.isActive)
+    const successfulCancellations = []
+    const unsuccessfulCancellations = []
+
+    for (let blockOrder of activeBlockOrders) {
+      try {
+        await this.cancelBlockOrder(blockOrder.id)
+        successfulCancellations.push(blockOrder.id)
+      } catch (e) {
+        unsuccessfulCancellations.push(blockOrder.id)
+      }
+    }
+
+    return { successfulCancellations, unsuccessfulCancellations }
+  }
+
+  /**
    * Get existing block orders
    * @param  {String} market to filter by
    * @return {Array<BlockOrder>}
--- a/broker-daemon/block-order-worker/index.js
+++ b/broker-daemon/block-order-worker/index.js
@@ -361,26 +361,29 @@ class BlockOrderWorker extends EventEmitter {
   /**
    * Cancel all active orders for a given market
    * @param  {string} market to cancel orders on
-   * @return {object} successfulCancellations and unsuccessfulCancellations as arrays of blockorderids
+   * @return {object} cancelledOrders and failedToCancelOrders as arrays of blockorderids
    */
   async cancelActiveOrders (market) {
     this.logger.info('Cancelling all active orders for market', { market })
 
     const blockOrders = await this.getBlockOrders(market)
     const activeBlockOrders = blockOrders.filter(blockOrder => blockOrder.isActive)
-    const successfulCancellations = []
-    const unsuccessfulCancellations = []
+    const cancelledOrders = []
+    const failedToCancelOrders = []
 
-    for (let blockOrder of activeBlockOrders) {
-      try {
-        await this.cancelBlockOrder(blockOrder.id)
-        successfulCancellations.push(blockOrder.id)
-      } catch (e) {
-        unsuccessfulCancellations.push(blockOrder.id)
-      }
-    }
+    await Promise.all(
+      activeBlockOrders.map(async (blockOrder) => {
+        try {
+          await this.cancelBlockOrder(blockOrder.id)
+          cancelledOrders.push(blockOrder.id)
+        } catch (e) {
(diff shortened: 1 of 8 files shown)


martineehrlich:
[Cancel orders when releasing channels](https://trello.com/c/ieLGxqrG/1131-cancel-orders-when-releasing-channels)

martineehrlich:
I still need to manually test and I am thinking that the naming for successfulCancellations and unsuccessfulCancellations could change.

Also I am aware that getBlockOrders finds the blockorders, and then I call cancelBlockOrder with the blockOrder.id, which gets the record from the database again. It didn't really make sense to refactor the cancelBlockOrder code which requires an id for the sake of not double fetching on a call that will never be huge anyways. 

Also, for releaseChannels, I wasn't sure what we wanted the behavior to be. Currently I am just logging because I wasn't sure if it made sense to pass error messages about order cancellation to the user on channel closure.

broker-daemon/block-order-worker/index.js
```diff
@@ -358,6 +358,31 @@ class BlockOrderWorker extends EventEmitter {
...
+    this.logger.info('Cancelling all active orders for market', { market })
+
+    const blockOrders = await this.getBlockOrders(market)
+    const activeBlockOrders = blockOrders.filter(blockOrder => blockOrder.isActive)
+    const successfulCancellations = []
+    const unsuccessfulCancellations = []
+
+    for (let blockOrder of activeBlockOrders) {
+      try {
+        await this.cancelBlockOrder(blockOrder.id)
```
treygriffith: having this `await` in the loop will cause each of these to cancel in parallel, which seems unnecessarily slow.

broker-daemon/block-order-worker/index.js
```diff
@@ -358,6 +358,31 @@ class BlockOrderWorker extends EventEmitter {
...
+        successfulCancellations.push(blockOrder.id)
+      } catch (e) {
+        unsuccessfulCancellations.push(blockOrder.id)
```
treygriffith: I think we need to capture the error here somehow, either by returning it or logging it

broker-daemon/block-order-worker/index.spec.js
```diff
@@ -1211,6 +1211,73 @@ describe('BlockOrderWorker', () => {
     })
   })
 
+  describe.only('#cancelActiveOrders', () => {
```
treygriffith: make sure to lose this only before merging

broker-daemon/broker-rpc/wallet-service/release-channels.js
```diff
@@ -84,6 +84,12 @@ async function releaseChannels ({ params, logger, engines, orderbooks }, { Relea
     closeChannels(counterEngine, counterSymbol, force, logger)
   ])
 
+  if (base.status === RELEASE_STATE.RELEASED && counter.status === RELEASE_STATE.RELEASED) {
+    const { successfulCancellations, unsuccessfulCancellations } = await blockOrderWorker.cancelActiveOrders(market)
```
treygriffith: Probably want this to happen before, or in tandem with the channels closing. One of the channels failing to close resulting in the orders not being cancelled is a bit of weird behavior I think.

broker-daemon/block-order-worker/index.js
```diff
@@ -358,6 +358,34 @@ class BlockOrderWorker extends EventEmitter {
...
+    const cancelledOrders = []
+    const failedToCancelOrders = []
+
+    await Promise.all(
+      activeBlockOrders.map(async (blockOrder) => {
+        try {
+          await this.cancelBlockOrder(blockOrder.id)
+          cancelledOrders.push(blockOrder.id)
+        } catch (e) {
+          this.logger.error('Failed to cancel order', e)
```
treygriffith: probably need the blockorder id in the logger. Also - if you pass the error as the second parameter to the logger it doesn't work well, it's better like:

```
this.logger.error('Failed to cancel block order', { blockOrderId: blockOrder.id, error: e })
```

broker-daemon/broker-rpc/wallet-service/release-channels.js
```diff
@@ -67,6 +67,10 @@ async function releaseChannels ({ params, logger, engines, orderbooks }, { Relea
     throw new PublicError(`${market} is not being tracked as a market.`)
   }
 
+  const { cancelledOrders, failedToCancelOrders } = await blockOrderWorker.cancelActiveOrders(market)
+  logger.info('Successfully cancelled orders', { orders: cancelledOrders })
+  logger.info('Failed to cancel orders', { orders: failedToCancelOrders })
```
treygriffith: weird to have this log if no block orders failed to cancel

broker-daemon/block-order-worker/index.js
```diff
@@ -358,6 +358,34 @@ class BlockOrderWorker extends EventEmitter {
     return blockOrder
   }
 
+  /**
+   * Cancel all active orders for a given market
+   * @param  {string} market to cancel orders on
+   * @return {object} cancelledOrders and failedToCancelOrders as arrays of blockorderids
```
dannypaz: For JSDoc, the representation of your return value should look like:

```
@returns {object} result
@returns {Array<string>} result.cancelledOrders ...
etc.. etc..
```

broker-daemon/block-order-worker/index.js
```diff
@@ -358,6 +358,36 @@ class BlockOrderWorker extends EventEmitter {
...
+          await this.cancelBlockOrder(blockOrder.id)
+          cancelledOrders.push(blockOrder.id)
+        } catch (e) {
+          this.logger.error('Failed to cancel block order', { blockOrderId: blockOrder.id, error: e })
+          failedToCancelOrders.push(blockOrder.id)
+        }
+      })
+    )
+
+    return { cancelledOrders, failedToCancelOrders }
```
treygriffith: [nit] might be nice to log the number of cancelled orders (and failed)

PR #411, opened 2019-02-15
add cancel all orders command

martineehrlich:
## Description
A user should be able to cancel all their orders for a market instead of having to go through and cancel one by one. 

## Related PRs
https://github.com/sparkswap/broker/pull/408

## Todos
- [x] Tests
- [x] Documentation
- [x] Link to Trello


broker-cli/commands/order/cancel-all.js
```diff
@@ -0,0 +1,33 @@
...
+  } = opts
+
+  const request = {
+    market
+  }
+
+  try {
+    const client = new BrokerDaemonClient(rpcAddress)
+    await client.orderService.cancelAllBlockOrders(request)
+    logger.info(`Cancelled all orders on ${market} market`)
```
treygriffith: probably worth noting how many block orders were cancelled, and if any failed to cancel we should direct the user to check the logs.

broker-daemon/proto/broker.proto
```diff
@@ -695,6 +713,23 @@ service OrderService {
...
+   * orderService.cancelAllBlockOrders({ market: 'BTC/LTC' }, { deadline }, (err) => {
+   *   if (err) return console.error(err)
+   * })
+   * ```
+   *
+   * ```shell
+   * > sparkswap order cancel-all --market btc/ltc
+   * ```
+   *
+   * > The above command will cancel the block order, but has no output.
```
treygriffith: there actually is output in the current shell implementation

martineehrlich:
[cancel all open orders](https://trello.com/c/frRa7dJM/347-cancel-all-open-orders)

broker-cli/commands/order/cancel-all.js
```diff
@@ -0,0 +1,33 @@
...
+  const {
+    rpcAddress,
+    market
+  } = opts
+
+  const request = {
+    market
+  }
+
+  try {
```
dannypaz: @martineehrlich per our talk, we should add a prompt that asks the user if they _really_ want to cancel all their orders

broker-daemon/broker-rpc/order-service/cancel-all-block-orders.js
```diff
@@ -0,0 +1,20 @@
...
+ * @param {GrpcUnaryMethod~request} request - request object
+ * @param {Object} request.params - Request parameters from the client
+ * @param {BlockOrderWorker} request.blockOrderWorker
+ * @param {Object} request.logger
+ * @return {Object}
+ */
+async function cancelAllBlockOrders ({ params, logger, blockOrderWorker }) {
+  const {
+    market
+  } = params
```
dannypaz: make this single line

broker-daemon/broker-rpc/order-service/cancel-all-block-orders.js
```diff
@@ -0,0 +1,20 @@
...
+
+  const { cancelledOrders, failedToCancelOrders } = await blockOrderWorker.cancelActiveOrders(market)
```
dannypaz: make this a double 🍔

broker-cli/commands/order/cancel-all.js
```diff
@@ -21,10 +28,23 @@ async function cancelAll (args, opts, logger) {
...
   try {
     const client = new BrokerDaemonClient(rpcAddress)
-    await client.orderService.cancelAllBlockOrders(request)
-    logger.info(`Cancelled all orders on ${market} market`)
+    const {
+      cancelledOrders = [],
+      failedToCancelOrders = []
+    } = await client.orderService.cancelAllBlockOrders(request)
+
+    var errorMessage = `Succesfully cancelled ${cancelledOrders.length} orders on ${market} market.`
```
dannypaz: use `let`

broker-cli/commands/order/cancel-all.js
```diff
@@ -1,5 +1,12 @@
 const BrokerDaemonClient = require('../../broker-daemon-client')
-const { handleError } = require('../../utils')
+const { handleError, askQuestion } = require('../../utils')
```
dannypaz: make it a double 🍔

broker-cli/commands/order/cancel-all.js
treygriffith: if we are going to have an interactive prompt, I think we also need a command line option to override it, e.g. `--force`. Otherwise you have to use it interactively, which is not great.

PR #412, opened 2019-02-16
open the channels that are missing when committing

treygriffith:
## Description
This PR changes the behavior of commit to open a second channel if only one had previously been opened.

It retains the behavior of doing nothing if either of the channels is smaller than the requested size.

## Todos
- [x] Tests


martineehrlich:
`This PR changes the behavior of commit to open a second channel if only one had previously been opened.` 

By second channel do you mean the channel from the relayer to the broker?

treygriffith:
> `This PR changes the behavior of commit to open a second channel if only one had previously been opened.`
> 
> By second channel do you mean the channel from the relayer to the broker?

This PR tries to open whatever channel is not currently open. Typically we would expect that to be the inbound channel from the relayer to the broker, but this change handles both cases.

PR #415, opened 2019-02-21
Fix/regtest local

dannypaz:
## Description
This PR refactors some of the regtest work on the broker to allow for an `npm run build-local` command for local dev only.

TLS certificates from the Relayer now use Lets Encrypt (supported by grpc) instead of a dedicated cert server. This dedicated cert server is only available on local setups.

## Related PRs

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Link to Trello

### DIFF
--- a/.env-regtest-local-sample
+++ b/.env-regtest-local-sample
@@ -32,6 +32,9 @@ BTC_RPC_PASS=sparkswapbtc
 # Transaction Broadcast Address
 # BTC_ZMQPUBRAWTX=your.bitcoin.node.transaction.address
 
+# Regtest Host
+BITCOIND_CONNECT_HOST=host.docker.internal
+
 # Litecoin Settings
 #
 # We have provided default username/passwords for all active engines (ltc, btc)
@@ -47,6 +50,9 @@ LTC_RPC_PASS=sparkswapltc
 
 # Transaction Broadcast Address
 # LTC_ZMQPUBRAWTX=your.litecoin.node.transaction.address
+
+# Regtest Host
+LITECOIND_CONNECT_HOST=host.docker.internal
 
 # External IP Addresses
 #
--- a/.env-regtest-local-sample
+++ b/.env-regtest-local-sample
@@ -69,4 +69,4 @@ SECURE_PATH=~/.sparkswap/secure
 # be applied and used
 #
 COMPOSE_PATH_SEPARATOR=:
-COMPOSE_FILE=docker-compose.yml:docker-compose.bitcoind.yml:docker-compose.litecoind.yml:docker-compose.override.yml
+COMPOSE_FILE=docker-compose.yml:docker-compose.bitcoind.yml:docker-compose.litecoind.yml:docker-compose.regtest.yml:docker-compose.local.yml
--- a/.env-regtest-local-sample
+++ b/.env-regtest-local-sample
@@ -1,5 +1,5 @@
 ###############################################################
-# TestNet Sparkswap Environment File
+# Sparkswap Environment File
 #
 # These settings control the underlying functionality of sparkswapd and the
 # associated lightning implementations (engines).
@@ -7,14 +7,14 @@
 ###############################################################
 
 # Set the network that sparkswap and engines will work on.
-NETWORK=testnet
+NETWORK=regtest
 
 # Set secure credentials for your sparkswap daemon
 RPC_USER=sparkswap
 RPC_PASS=sparkswap
 
-# Relayer Host Addresses for TestNet
-RELAYER_RPC_HOST=relayer.testnet.sparkswap.com:28492
+# Relayer Host Addresses for RegTest
+RELAYER_RPC_HOST=host.docker.internal:28492
 
 # Bitcoin Settings
 #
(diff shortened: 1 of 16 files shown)


docker-compose.local.yml
```diff
@@ -15,15 +15,7 @@ services:
       - './:/home/app/'
       - '/home/app/node_modules'
       - '/home/app/broker-cli/node_modules'
+    # environment:
```
treygriffith: why is this in here if it's commented out?

docker-compose.regtest.yml
```diff
@@ -0,0 +1,21 @@
...
+# 1. copy file to `docker-compose.override.yml`
+# 2. restart all containers with `docker-compose up -d`
+##########################################
+
+version: '2.4'
+
+services:
+  bitcoind:
+    environment:
+      - CONNECT_HOST=host.docker.internal
```
treygriffith: Shouldn't this be the bitcoind node that is hosting the regtest network?

scripts/build.sh
```diff
@@ -167,3 +172,12 @@ if [ -f docker-compose.override.yml ]; then
   echo "WARNING: This may add unwanted settings to the broker that could affect how your daemon runs."
   echo ""
 fi
+
+if [ "$LOCAL" == "true" ]; then
+  echo "Downloading Local Relayer Cert..."
```
treygriffith: I like this change, it makes the local development much more predictable.

scripts/build.sh
```diff
@@ -167,3 +172,12 @@ if [ -f docker-compose.override.yml ]; then
...
+  # the path of this output is directly related to the SECURE_PATH that is set
+  # in the .env file
+  curl --silent -S --output ~/.sparkswap/secure/relayer-root.pem http://localhost:8080/cert
```
treygriffith: we should consider renaming this relayer pem to something more obviously local/regtest.

docker-compose.local.yml
```diff
@@ -15,15 +11,7 @@ services:
       - './:/home/app/'
       - '/home/app/node_modules'
       - '/home/app/broker-cli/node_modules'
+    environment:
+      - NODE_ENV=development
```
treygriffith: hmm, so does this mean that hosted regtest is considered a "production" environment? I don't think I agree with that characterization.

scripts/build.sh
```diff
@@ -167,3 +172,12 @@ if [ -f docker-compose.override.yml ]; then
   echo "WARNING: This may add unwanted settings to the broker that could affect how your daemon runs."
   echo ""
 fi
+
+if [ "$LOCAL" == "true" ]; then
+  echo "Downloading Local Relayer Cert..."
+  # the path of this output is directly related to the SECURE_PATH that is set
+  # in the .env file
+  curl --silent -S --output ~/.sparkswap/secure/relayer-root-regtest-local.pem http://localhost:8080/cert
```
treygriffith: so no more relayer cert host? We just assume that if you're in a local set up the Relayer is on localhost? (That might be reasonable, just want to make sure I understand the choice we're making here).

.env-regtest-sample
```diff
@@ -70,4 +75,4 @@ SECURE_PATH=~/.sparkswap/secure
 # be applied and used
 #
 COMPOSE_PATH_SEPARATOR=:
-COMPOSE_FILE=docker-compose.yml:docker-compose.bitcoind.yml:docker-compose.litecoind.yml:docker-compose.override.yml
+COMPOSE_FILE=docker-compose.yml:docker-compose.bitcoind.yml:docker-compose.litecoind.yml:docker-compose.regtest.yml
```
treygriffith: so no override file at all? Shouldn't we add it to the end of the path here?

.env-regtest-local-sample
```diff
@@ -0,0 +1,78 @@
...
+# Path to rpc certs and identity keys
+SECURE_PATH=~/.sparkswap/secure
+
+# Docker compose files
+# docker-compose by default will look for the docker-compose.yml file and apply the docker-compose.override.yml file if
+# it exists. If we specify the COMPOSE_FILE variable, all compose files joined by the COMPOSE_PATH_SEPARATOR will
+# be applied and used
+#
+COMPOSE_PATH_SEPARATOR=:
+COMPOSE_FILE=docker-compose.yml:docker-compose.bitcoind.yml:docker-compose.litecoind.yml:docker-compose.regtest.yml:docker-compose.local.yml
```
treygriffith: shouldn't we have an override file at the end of this?

scripts/build.sh
dannypaz: Yup. FWIW the RELAYER_CERT_HOST was never used outside of downloading the cert

docker-compose.local.yml
dannypaz: @treygriffith we would consider hosted regtest to be `staging`. development is local

dannypaz: I would not consider hosted regtest production either

.env-regtest-sample
dannypaz: @treygriffith we no longer require an override file for our dev setup. If we add `docker-compose.override.yml` to this file AND we dont have an override file, youll receive an error:

```
ERROR: .FileNotFoundError: [Errno 2] No such file or directory: './docker-compose.override.yml'
```

.env-regtest-local-sample
dannypaz: see https://github.com/sparkswap/broker/pull/415#discussion_r260016215

docker-compose.local.yml
dannypaz: A background on this flag, this is a solution for broker configuration so that we can have regtest on a staging and local environment. I stuck to environment conventions with `NODE_ENV` cause it was easier than creating a separate `BROKER_ENVIRONMENT` flag that controlled this

PR #420, opened 2019-03-02
move convert balance and relayer channel opening logic to the relayer

martineehrlich:
## Description
When opening channels, we open an outbound channel to the relayer with the specified balance. Then we convert the balance to the inverseSymbol and call to the relayer to open a channel to the broker with that amount. 

We want this conversion to happen in the relayer and eventually make it dynamic based on prevailing market prices on Sparkswap or elsewhere.

This PR:
- moves conversion of outbound -> inbound amount to the relayer
- moves checks on whether or not we should open an inbound channel that currently exists in the commit call to the createChannel call on the relayer

## Related PRs

## Todos
- [x] Tests
- [x] Documentation
- [x] Link to Trello


broker-daemon/broker-rpc/wallet-service/commit.js
```diff
@@ -116,24 +96,21 @@ async function commit ({ params, relayer, logger, engines, orderbooks }, { Empty
...
-    }
-  } else {
-    logger.debug('Inbound channel already exists', { balance: maxInboundBalance })
+  const paymentChannelNetworkAddress = await inverseEngine.getPaymentChannelNetworkAddress()
+
+  logger.debug('Creating inbound channel', { balance, symbol, inverseSymbol })
+
+  try {
+    const authorization = relayer.identity.authorize()
+    await relayer.paymentChannelNetworkService.createChannel({
```
treygriffith: maybe `createInboundChannel` ? Not sure that it's worth the rename, but it seems like an apt description given the other renames you did.

dannypaz:
[When creating a channel on the relayer from the broker in commitBalance, get rid of hardcoding BTC/LTC conversion](https://trello.com/c/lVI4l0E1/465-when-creating-a-channel-on-the-relayer-from-the-broker-in-commitbalance-get-rid-of-hardcoding-btc-ltc-conversion)

PR #426, opened 2019-03-08
FIX - CLI stdin pipe issue

dannypaz:
## Description
This PR fixes 2 issues with the CLI that prevented the CLI from being used programmatically when interactive prompts were activated

Right now, users are unable to effectively pipe information into the sparkswap wallet commands that involve multiple questions: `echo "mypassword\nmypassword\n" | sparkswap wallet create btc`

**Tasks**
1. Created `ask-password` method that provides input suppression for multi-message questions
2. Modified suppression for `ask-question`

**EXAMPLES**

OLD: Sensitive output is _not_ suppressed when using pipes

screen shot 2019-03-07 at 6 45 04 pm

OLD: Multi-line sensitive output is not suppressed and does not work correctly
screen shot 2019-03-07 at 6 46 22 pm

NEW: Sensitive output is suppressed w/ pipes
screen shot 2019-03-07 at 6 45 08 pm

NEW: User can use pipes w/ password prompts
screen shot 2019-03-07 at 6 47 46 pm

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Link to Trello


broker-cli/utils/ask-password.js
```diff
@@ -0,0 +1,59 @@
...
+        })
+
+        // Suppress output when the user confirms their password
+        rl.silent = true
+      })
+
+      // Suppress output when the user enters their password
+      rl.silent = true
+    } catch (e) {
+      rl.close()
```
martineehrlich: Can you write a test for this case?

PR #427, opened 2019-03-08
FEATURE - removed process.env.NODE_ENV for global setting

dannypaz:
## Description
This PR adds a global var `global.sparkswap.env` to replace `process.env` usage for the broker.

Configuration works as follows:
1. .env is loaded
2. variables in docker-compose are loaded
3. all variables from the docker container are input into sparkswapd via `start-sparkswapd.sh`
4. default configuration is loaded from `config.json` in broker daemon

We've decided to omit the usage of environment variables from within the application to have a single point of configuration into the app to avoid conflicts with other programs/software

## Related PRs

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Link to Trello

### DIFF
--- a/broker-cli/utils/validations.js
+++ b/broker-cli/utils/validations.js
@@ -21,6 +21,16 @@ const BLOCKCHAIN_NETWORKS = Object.freeze([
 ])
 
 /**
+ * @constant
+ * @type {Object}
+ * @default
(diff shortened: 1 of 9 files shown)


dannypaz:
Closing this as its not a high priority right now

PR #431, opened 2019-03-11
add check for orphans script on broker-daemon

martineehrlich:
## Description
We want to detect unused modules because we currently have files required through index files and not directly. We have had issues in the past where we didn't add the modules to the index file. We have already added this on the relayer.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Link to Trello


scripts/check-unused.js
```diff
@@ -0,0 +1,38 @@
+const madge = require('madge')
+require('colors')
+
+/**
+ * Regexp string to exclude spec files from orphans
+ * @type {string}
+ * @constant
+ */
+const REGEXP_TO_EXCLUDE = '.spec.js$'
```
treygriffith: might be nice to call this something more descriptive, like `TEST_FILE_REGEXP`

PR #433, opened 2019-03-14
Broker re-places orders when gRPC stream is unavailable

bradleystachurski:
## Description
When the broker loses a connection with the relayer, the broker currently moves all active orders to a failed state. This PR adds logic to attempt to re-place the orders for a period of time (up to 5 minutes using 10 second intervals). If the relayer is still unavailable after this time period, the broker then moves the order to a failed state.

## Related PRs

## Todos
- [x] Tests
- [x] Documentation
- [x] Link to Trello


broker-daemon/block-order-worker/index.js
```diff
@@ -573,7 +574,15 @@ class BlockOrderWorker extends EventEmitter {
     // reject the entire block order if an underlying order fails
     osm.once('reject', async () => {
       try {
-        await this.failBlockOrder(blockOrder.id, osm.order.error)
+        if (osm.shouldRetry()) {
+          const validation = async () => {
```
dannypaz: probably have to rename this function to something more descriptive about what we're trying to do.

broker-daemon/block-order-worker/index.js
```diff
@@ -573,7 +574,15 @@ class BlockOrderWorker extends EventEmitter {
...
+            await this.relayer.adminService.healthCheck({})
```
treygriffith: feels a little weird to have this in here. is there something more abstract we can use? like something on the top level of block order worker that is like `relayerIsAvailable`?

broker-daemon/block-order-worker/index.js
```diff
@@ -573,7 +574,15 @@ class BlockOrderWorker extends EventEmitter {
...
+            await this.workBlockOrder(blockOrder, Big(osm.order.baseAmount))
+          }
+          await exponentialBackoff(validation, 10)
```
treygriffith: don't you want to retry the `shouldRetry` check as well?

treygriffith: does `workBlockOrder` check that this block order hasn't been cancelled or failed?

broker-daemon/state-machines/order-state-machine.js
```diff
@@ -280,7 +289,8 @@ const OrderStateMachine = StateMachine.factory({
       }
 
       const errHandler = (e) => {
-        this.reject(e)
+        const relayerError = new Error(ORDER_ERROR_CODES.RELAYER_UNAVAILABLE)
```
treygriffith: Is this the only error that's possible? What's in the `e` parameter here?

bradleystachurski: Good question. When the gRPC stream ends from with the relayer it sends an error code 14, unavailable. Added a conditional to only add the `relayerError` when the broker receives code 14.

broker-daemon/block-order-worker/index.js
bradleystachurski: We should be okay kicking things off with the `shouldRetry` without having to check if it should retry each iteration. The broker would have reestablished a gRPC stream with the relayer and would have replaced the block order if the order is still active.

`workBlockOrder` does check if the block order is in a workable state (i.e. if the order is active).

bradleystachurski:
https://trello.com/c/EyhaVmbL/901-re-place-orders-that-are-cancelled-by-the-relayer-going-down

broker-daemon/block-order-worker/index.spec.js
```diff
@@ -2115,22 +2118,77 @@ describe('BlockOrderWorker', () => {
...
+
         rejectListener = onceStub.withArgs('reject').args[0][1]
       })
 
       it('registers an event on an order for order rejection', async () => {
         expect(onceStub).to.have.been.calledWith('reject', sinon.match.func)
       })
 
-      it('fails a block order if the call is rejected', async () => {
+      it('retries the block order when it should', async () => {
```
dannypaz: would add the situation here to be more specific. Example `retries the block order when the relayer is available`

broker-daemon/state-machines/order-state-machine.js
```diff
@@ -280,7 +289,13 @@ const OrderStateMachine = StateMachine.factory({
       }
 
       const errHandler = (e) => {
-        this.reject(e)
+        // code 14 corresponds to gRPC UNAVAILABLE
+        if (e.code === 14) {
```
dannypaz: totally forgot about this, but we can actually use the grpc library for this. `grpc.status.UNAVAILABLE` should make to code 14 instead of using the number.

Code would look like `if (e.code === grpc.status.UNAVAILABLE)`

dannypaz:
[Re-place orders that are cancelled by the Relayer going down](https://trello.com/c/EyhaVmbL/901-re-place-orders-that-are-cancelled-by-the-relayer-going-down)

bradleystachurski:
[Re-place orders that are cancelled by the Relayer going down](https://trello.com/c/EyhaVmbL/901-re-place-orders-that-are-cancelled-by-the-relayer-going-down)

broker-daemon/utils/retry.js
```diff
@@ -0,0 +1,53 @@
...
+ */
+async function retry (callFunction, message, attempts = RETRY_ATTEMPTS, delayTime = DELAY) {
+  try {
+    var res = await callFunction()
+  } catch (error) {
+    if (attempts > 0) {
+      const attemptsLeft = attempts - 1
+
+      if (message) {
+        logger.error(message, { delayTime: delayTime / 1000, attemptsLeft })
```
treygriffith: [nit] you should round `delayTime / 1000` so the logs don't get spammed with long decimal places

broker-daemon/utils/retry.js
```diff
@@ -0,0 +1,53 @@
...
+      } else {
+        logger.error(`Error calling ${callFunction}. Retrying in ${delayTime / 1000} seconds, attempts left: ${attemptsLeft}`)
```
treygriffith: what does logging `callFunction` like this do? I would expect that it logs the entire body of the function when you probably just want its name.

broker-daemon/utils/retry.spec.js
```diff
@@ -0,0 +1,88 @@
...
+    retry.__set__('delay', delay)
+  })
+
+  it('calls the call function', async () => {
+    await retry(callFunction, message, attempts, delayTime)
+
+    expect(callFunction).to.be.to.have.been.calledOnce()
+  })
+
+  it('returns the response of the callFunction if there were no errors', async () => {
```
treygriffith: Probably worth it to make sure that `callFunction` is called only once if it has no errors on the first time

broker-daemon/block-order-worker/index.js
```diff
@@ -570,13 +585,39 @@ class BlockOrderWorker extends EventEmitter {
...
+    // try to re-place the block order if the relayer drops connection, while allowing the broker to
+    // update block order status (i.e. broker can still cancel orders if relayer goes down)
     osm.once('reject', async () => {
+      if (osm.shouldRetry()) {
+        const retryBlockOrder = async () => {
+          if (!await this.relayerIsAvailable()) {
+            throw new Error('Relayer not available')
+          }
+
+          this.logger.info('Retrying order for block order', { order: osm.order })
```
treygriffith: probably want this to be the order id, right? not the whole order?

broker-daemon/block-order-worker/index.js
```diff
@@ -570,13 +585,39 @@ class BlockOrderWorker extends EventEmitter {
...
+
+          // Get the existing order in case an update to order status happened between retries
+          const updatedBlockOrder = await this.getBlockOrder(blockOrder.id)
```
treygriffith: hmm, what's going on here? Shouldn't the block order be passed by reference throughout the worker? What kind of update might happen that would cause us to pull from the data store?

broker-daemon/block-order-worker/index.js
```diff
@@ -570,13 +585,39 @@ class BlockOrderWorker extends EventEmitter {
...
+          await this.workBlockOrder(updatedBlockOrder, Big(osm.order.baseAmount))
```
treygriffith: do we need to convert the base amount to a `Big` here?

broker-daemon/block-order-worker/index.js
```diff
@@ -570,13 +585,39 @@ class BlockOrderWorker extends EventEmitter {
...
+        }
+
+        // This will try to rework each order within a block order every 10 seconds for 5 minutes
+        // We return the retry function so we can fail the block order if the retry logic fails
+        try {
+          return retry(retryBlockOrder, 'Error reworking block order', RETRY_ATTEMPTS, DELAY)
```
treygriffith: [nit] would be nice to include the block order id in the error message

broker-daemon/block-order-worker/index.js
```diff
@@ -570,13 +585,39 @@ class BlockOrderWorker extends EventEmitter {
...
+        } catch (e) {
+          // If retry has failed, we log and fall through to fail the block order below
+          this.logger.error('Retrying block order has failed. Attempting to change order status to failed.')
```
treygriffith: [nit] same comment here about block order id and perhaps the underlying order id

broker-daemon/block-order-worker/index.js
```diff
@@ -570,13 +585,39 @@ class BlockOrderWorker extends EventEmitter {
...
+      }
+
+      // This handles failing block orders if retrying has failed or the osm shouldn't retry
       try {
         await this.failBlockOrder(blockOrder.id, osm.order.error)
       } catch (e) {
         this.logger.error(`BlockOrder failed on setting a failed status from order`, { id: blockOrder.id, error: e.stack })
       }
+
       osm.removeAllListeners()
```
treygriffith: do we need to change when we remove listeners based on retrying? are there cases where keeping these listeners alive will cause problems while we're re-trying?

broker-daemon/block-order-worker/index.js
```diff
@@ -790,6 +831,21 @@ class BlockOrderWorker extends EventEmitter {
...
+
+  /**
+   * Checks if the relayer is available by pinging the healthCheck endpoint. If available,
+   * returns true. False otherwise.
+   * @private
+   * @returns {Promise<boolean>}
+   */
+  async relayerIsAvailable () {
+    try {
+      await this.relayer.adminService.healthCheck({})
```
treygriffith: is the only output here a return value or a `throw`? Is there no value that could be returned that would cause us to determine that the Relayer was unavailable?

broker-daemon/block-order-worker/index.spec.js
```diff
@@ -2115,22 +2118,94 @@ describe('BlockOrderWorker', () => {
...
+        it('does not try re-placing the block order when relayer is not available', async () => {
+          relayerIsAvailableStub.resolves(false)
+
+          await rejectListener()
+          const retryStub = retry.args[0][0]
+          expect(retryStub()).to.have.been.rejectedWith('not available')
+        })
+      })
+
+      context('should NOT retry', () => {
```
treygriffith: make sure to have a test showing that the block order was NOT re-worked in this case

broker-daemon/state-machines/order-state-machine.js
```diff
@@ -280,7 +290,12 @@ const OrderStateMachine = StateMachine.factory({
       }
 
       const errHandler = (e) => {
-        this.reject(e)
+        if (e.code === grpc.status.UNAVAILABLE) {
+          const relayerError = new Error(ORDER_ERROR_CODES.RELAYER_UNAVAILABLE)
+          this.reject(relayerError)
```
treygriffith: Could consolidate this and the lower line with something like:

```
const relayerError = e.code === grpc.status.UNAVAILABLE ? new Error(ORDER_ERROR_CODES.RELAYER_UNAVAILABLE) : e
this.reject(relayerError)
```

broker-daemon/state-machines/order-state-machine.js
```diff
@@ -346,6 +361,15 @@ const OrderStateMachine = StateMachine.factory({
       this.triggerComplete()
     },
 
+    /**
+     * Returns true if there is a relayer error associated with the order, false if not
+     * This is just a getter function, no transition associated
+     * @returns {boolean}
+     */
+    shouldRetry: function () {
+      return !!this.order.error && this.order.error.message === ORDER_ERROR_CODES.RELAYER_UNAVAILABLE
```
treygriffith: why use the `!!` instead of just `this.order.error && ...` ?

broker-daemon/block-order-worker/index.js
bradleystachurski: Yup, `workBlockOrder` is expecting target depth to be passed in as `Big`

PR #436, opened 2019-03-22
run uncommitted balance checks only when there is not already an outbound channel open

martineehrlich:
## Description
When a user opens channels, and the outbound succeeds but the inbound fails, we made it so they can commit again and we will just bypass creating the outbound channel. However, we still have checks in place that if they have an uncommitted balance lower than the channel they are trying to create, we do not allow them to commit funds. We need to run this check only if there is not already an outbound channel open.

## Related PRs

## Todos
- [x] Tests
- [x] Documentation
- [x] Link to Trello

### DIFF
--- a/broker-daemon/broker-rpc/wallet-service/commit.js
+++ b/broker-daemon/broker-rpc/wallet-service/commit.js
@@ -89,7 +89,10 @@ async function commit ({ params, relayer, logger, engines, orderbooks }, { Empty
 
     if ((Big(balance).plus(engine.feeEstimate)).gt(totalUncommittedBalance)) {
       const uncommittedCommon = totalUncommittedBalance.div(engine.quantumsPerCommon)
-      throw new Error(`Amount specified is larger than your current uncommitted balance of ${uncommittedCommon.toString()} ${symbol}`)
+      const feeEstimateCommon = Big(engine.feeEstimate).div(engine.quantumsPerCommon)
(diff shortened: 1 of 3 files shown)


martineehrlich:
[CLI errors when re-committing the same amount](https://trello.com/c/Xt6PqnnO/1297-cli-errors-when-re-committing-the-same-amount)

broker-daemon/broker-rpc/wallet-service/commit.js
```diff
@@ -84,6 +84,18 @@ async function commit ({ params, relayer, logger, engines, orderbooks }, { Empty
   }
 
   if (!maxOutboundBalance) {
+    const uncommittedBalance = await engine.getUncommittedBalance()
+    const totalUncommittedBalance = Big(uncommittedBalance)
+
+    if (totalUncommittedBalance.eq(0)) {
+      throw new Error('Your current uncommitted balance is 0, please add funds to your daemon')
```
treygriffith: Isn't this a duplicative check to the one below?

broker-daemon/broker-rpc/wallet-service/commit.js
```diff
@@ -84,6 +84,18 @@ async function commit ({ params, relayer, logger, engines, orderbooks }, { Empty
...
+    }
+
+    if (Big(balance).gt(totalUncommittedBalance)) {
```
treygriffith: we should probably use the fee estimate here as well, right?

PR #437, opened 2019-03-23
Better help for market name in CLI

treygriffith:
## Description
Caporal doesn't support subcommands well, in particular required options for a subcommand just shows its validation message which can be quite opaque.

This change make sure `marketName` is marked as required throughout the CLI and shows a better error message when it is left out of subcommands to improve usability.


broker-cli/commands/market/index.js
```diff
@@ -53,10 +53,10 @@ module.exports = (program) => {
     .command(`market ${SUPPORTED_COMMANDS.SUPPORTED_MARKETS}`, 'Get the markets currently supported')
     .option('--rpc-address [rpc-address]', RPC_ADDRESS_HELP_STRING, validations.isHost)
     .command(`market ${SUPPORTED_COMMANDS.MARKET_STATS}`, 'Get statistics (price ticker information) for a particular market for a period of the last 24 hours')
-    .option('--market [marketName]', MARKET_NAME_HELP_STRING, validations.isMarketName)
+    .option('--market <marketName>', MARKET_NAME_HELP_STRING, validations.isMarketName, null, true)
```
dannypaz: @treygriffith +1

broker-cli/commands/order/index.js
```diff
@@ -61,17 +61,17 @@ module.exports = (program) => {
       }
     })
     .command(`order ${SUPPORTED_COMMANDS.SUMMARY}`, 'View your orders')
+    .option('--market <marketName>', MARKET_NAME_HELP_STRING, validations.isMarketName, null, true)
```
dannypaz: just an FYI. None of these validations or defaults are used (they dont affect the sub commands because of the way we've rigged it). We need to actually run the marketname validation in the `.action` handler for subcommands

broker-cli/utils/validations.js
```diff
@@ -65,9 +73,9 @@ function isMarketName (str) {
       return str.toUpperCase()
     }
 
-    throw new Error()
+    throw new Error('[empty error to trigger the catch]')
   } catch (e) {
-    throw new Error('Market Name format is incorrect')
+    throw new ValidationError(`Market Name should be specified with '--market <marketName>', where <marketName> is two symbols separated by a '/' e.g. 'BTC/LTC'`)
```
dannypaz: Is there a name for this format? (maybe official terminology from investopedia)

dannypaz: Reason why I ask is because of the explanation in this error

bradleystachurski: I'm not aware of any official designation from traditional finance, but usually hear it referred to as 'trading pair'. `base` and `quote` should be fairly ubiquitous across the industry, so this could be rephrased similar to:

`-- market <marketName>, where <marketName> is the trading pair designated as <baseToken>/<quoteToken>`

That said, I think the the error message as stated would be enough for the user to figure out the correct format.

broker-cli/commands/order/index.js
treygriffith: hmm. Do you think we should just remove them then? This PR adds a bunch of them back in, but if they aren't used it's a bit strange.

PR #438, opened 2019-03-25
Feature - keepalive grpc

dannypaz:
## Description
This PR updates all gRPC clients and servers to send "KEEP_ALIVE" messages to prevent our grpc servers from throwing away idle connections on long-lived actions.

Official docs for keep alive on grpc: https://github.com/grpc/grpc/blob/master/doc/keepalive.md

**Example Issues**
1. `sparkswap orderbook --market btc/ltc` fails to send updates once the gRPC server on broker throws away the connection after the connection is broken (idle time is 2 hours on an uninteruppted connection)
2. Interchain router may timeout while waiting for swap to return
3. `watchMarket` between broker and relayer will timeout after a given time due to the server (relayer) thinking that our idle connection is bad and will disconnect WITHOUT letting the broker know that the connection is dead

## Related PRs

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Link to Trello


dannypaz:
[Broker healthcheck hangs if connection to grpc is timed out](https://trello.com/c/tSZy3xst/1295-broker-healthcheck-hangs-if-connection-to-grpc-is-timed-out)

PR #441, opened 2019-03-26
unfreeze grpc options objects

dannypaz:
## Description
I introduced a regression in <https://github.com/sparkswap/broker/pull/438>. I used `Object.freeze` on all grpc options, however grpc will try to modify the object and will fail because it cant.

This PR removes Object.freeze from all grpc options

## Related PRs

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Link to Trello


dannypaz:
[Bug: Broker fails when gRPC tries to mutate grpc options](https://trello.com/c/GU1m58MN/1344-bug-broker-fails-when-grpc-tries-to-mutate-grpc-options)

PR #442, opened 2019-03-26
Fix - Default deadlines for grpc clients

dannypaz:
## Description
This PR adds deadlines to all non-streaming calls by implementing `ClientInterceptors` that are available in grpc. This allows us to set global defaults per service instead of having to require `grpcDeadline` everywhere.

Things to note:
1. we cannot use grpc-caller to create a grpc client, instead we now use the `caller.wrap` static method to be able to promisfy all of our grpc calls
    - all streaming calls failed when using grpc-caller to create a grpc client
2. Duplication between broker-cli/broker-daemon was on purpose to keep them as separate as possible
3. This PR is actually introducing the usage of `grpcDeadline` (we previously had no cases that it was being used other than healthcheck)

## Related PRs

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Link to Trello


broker-cli/broker-daemon-client/index.js
```diff
@@ -58,6 +59,19 @@ const GRPC_STREAM_OPTIONS = {
...
 
+/**
+ * gRPC client interceptor that allows us to set a deadline for all calls in a
+ * service
+ *
+ * @param {Object} options
+ * @param {Function} nextCall
+ * @returns {grpc.InterceptingCall}
+ */
+function grpcInterceptor (options, nextCall) {
```
treygriffith: maybe put this into a util as well? `grpcDeadlineInterceptor` ?

broker-cli/broker-daemon-client/index.js
```diff
@@ -58,6 +59,19 @@ const GRPC_STREAM_OPTIONS = {
...
+  options.deadline = grpcDeadline()
```
treygriffith: I would only set the deadline if it's not already set, that way you can still specify a longer / shorter deadline if it's desired for a particular call.

broker-cli/utils/grpc-deadline.js
```diff
@@ -0,0 +1,24 @@
...
+ * @default
+ */
+const DEFAULT_TIMEOUT_IN_SECONDS = 5
+
+/**
+ * gRPC uses the term `deadline` which is a timeout feature that is an absolute
+ * point in time, instead of a duration.
+ *
+ * @param {number} [timeoutInSeconds=DEFAULT_TIMEOUT_IN_SECONDS]
+ * @returns {number} deadline in seconds
```
bradleystachurski: I think this should return `Date`

dannypaz:
[Add deadline to all grpc commands in cli](https://trello.com/c/ySLIA5Ji/1345-add-deadline-to-all-grpc-commands-in-cli)

broker-cli/utils/grpc-deadline.js
dannypaz: @bradleystachurski the type is actually a number

```
> typeof new Date().setSeconds(new Date().getSeconds() + 5)
'number'
```

bradleystachurski: Ahh, I misinterpreted the the docs on this method that it is `an absolute point in time` to refer to a `Date` object. The docs for grpc specify `deadline` (https://grpc.io/grpc/node/grpc.html#~Deadline) can be either `number | Date`, so disregard my comment.

dannypaz: @bradleystachurski No worries. I should have elaborated on the comment as well. 

Just to be thorough, we return a number here, but signify its use in the comment. We use this same pattern in a few places where we return a Date string (ISO8601) and say `@returns {string} date string in ISO8601 format`) 

However... if I was actually returning a Date object (`typeof new Date` === `object`) then I would expect that we use `@returns {Date}` since the return value has methods that are related to the Date object for JS

broker-daemon/relayer/relayer-client.js
```diff
@@ -86,11 +87,20 @@ class RelayerClient {
 
     this.credentials = channelCredentials
 
-    this.makerService = caller(this.address, this.proto.MakerService, this.credentials, GRPC_STREAM_OPTIONS)
-    this.takerService = caller(this.address, this.proto.TakerService, this.credentials, GRPC_STREAM_OPTIONS)
-    this.orderBookService = caller(this.address, this.proto.OrderBookService, this.credentials, GRPC_STREAM_OPTIONS)
-    this.paymentChannelNetworkService = caller(this.address, this.proto.PaymentChannelNetworkService, this.credentials)
-    this.adminService = caller(this.address, this.proto.AdminService, this.credentials)
+    const makerServiceClient = new this.proto.MakerService(this.address, this.credentials, GRPC_STREAM_OPTIONS)
+    this.makerServiceClient = caller.wrap(makerServiceClient, {}, { interceptors: [grpcDeadlineInterceptor] })
```
bradleystachurski: `this.makerServiceClient` can be renamed to `this.makerService`

dannypaz: Makes me worried that this passed our test specs. Good catch @bradleystachurski

broker-cli/utils/index.js
```diff
@@ -13,5 +15,7 @@ module.exports = {
   loadProto,
   Big,
   handleError,
-  basicAuth
+  basicAuth,
+  grpcDeadline,
```
treygriffith: do we actually want to export `grpcDeadline`, or just the interceptor?

broker-daemon/utils/grpc-deadline-interceptor.js
```diff
@@ -0,0 +1,35 @@
+const grpc = require('grpc')
+
+const grpcDeadline = require('./grpc-deadline')
```
treygriffith: I don't think this file exists (only the `broker-cli` one is in this PR I think)

dannypaz: @treygriffith we already had grpc-deadline available in the daemon so this file already exists on master

dannypaz:
@treygriffith RE: duplication. I felt the same way. Im not sure when we'll get to the split. The only file we have shared between the broker-cli and daemon is the CLI validations, everything else has been split

broker-cli/utils/index.js
dannypaz: we could probably remove it since its not used

PR #443, opened 2019-03-26
Fix - Update handleError for unregistered entity IDs

bradleystachurski:
## Description
When a user attempts to commit funds to the Relayer, if they haven't previously registered they will receive a cryptic error `No active entity exists for the provided Public Key`. This PR adds error handling for this case and prints a suggested help message to run `sparkswap register`.

image

## Todos
- [x] Tests
- [x] Documentation
- [x] Link to Trello

### DIFF
--- a/broker-cli/utils/error-handler.js
+++ b/broker-cli/utils/error-handler.js
@@ -2,6 +2,13 @@ require('colors')
 const { status } = require('grpc')
 
 /**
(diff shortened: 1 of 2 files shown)


bradleystachurski:
[Better error message when user hasn't registered with relayer](https://trello.com/c/fDBBsXon/1259-better-error-message-when-user-hasnt-registered-with-relayer)

PR #445, opened 2019-03-28
Feature/cli container

martineehrlich:
## Description
We want to be able to run the cli inside and outside of the container. To do this, the following changes are:
- add permanent config that gets copied (not mounted) inside of the container at the root directory. 
- install node_modules for cli in dockerfile
- add broker-cli/bin to path so we can just call sparkswap in container
- ensure we prepend / when we do not expand the file path to certs

## Related PRs

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Link to Trello


martineehrlich:
[Install broker-cli in the sparkswapd container](https://trello.com/c/np0uTLCB/1206-install-broker-cli-in-the-sparkswapd-container)

.sparkswap/config.js
```diff
@@ -0,0 +1,39 @@
+/**
+ * Sparkswap configuration for use inside the container. (Note the rpcCertPath is relative to being inside the container)
```
treygriffith: Can you rename this file to something more obvious, like `cli-container-config`, and not in hidden directory.

.sparkswap/config.js
```diff
@@ -0,0 +1,39 @@
...
+ * It will use the user defined rpc address, user and password if they exist, otherwise
+ * they will use the default.
+ */
+
+module.exports = {
+  /**
+   * Address of the host for the Broker Daemon gRPC Server
+   * @type {String}
+   */
+  rpcAddress: process.env.RPC_ADDRESS || 'localhost:27492',
```
treygriffith: `RPC_ADDRESS` will be something like `0.0.0.0:2749`, so should we just pull the port off here?

.sparkswap/config.js
```diff
@@ -0,0 +1,39 @@
...
+
+  /**
+   * Default path of the Broker Daemons RPC Public Cert
+   * @type {String}
+   */
+  rpcCertPath: '/secure/broker-rpc-tls.cert',
```
treygriffith: Worth adding a comment that it will always be here because this is where it is mounted in the container.

.sparkswap/config.js
```diff
@@ -0,0 +1,39 @@
...
+
+  /**
+   * Configuration for SSL between the CLI and Daemon. This setting is only required
+   * if you will be hosting the daemon remotely
+   * @type {Boolean}
+   */
+  disableAuth: false,
```
treygriffith: I would not include this config

.sparkswap/config.js
treygriffith: Worth noting that this config is for using the broker-cli inside the container specifically.

broker-cli/broker-daemon-client/index.js
```diff
@@ -80,7 +80,8 @@ class BrokerDaemonClient {
       if (certPathParts[0] === '~') {
         certPathParts[0] = os.homedir()
       }
-      const certPath = path.join(...certPathParts)
+
+      const certPath = path.normalize('/' + path.join(...certPathParts))
```
treygriffith: might be worth trying if there is a way to get the root path without this leading slash.

docker/sparkswapd/Dockerfile
```diff
@@ -39,6 +42,15 @@ COPY --from=builder /home/app/proto /home/app/proto
 
 # Copy node_modules
 COPY --from=builder /home/app/node_modules /home/app/node_modules
+COPY --from=builder /home/app/broker-cli/node_modules /home/app/broker-cli/node_modules
+
+ARG USER=root
```
treygriffith: I think this should be a path and the name should reflect that, e.g. `USER_HOME=/root`

broker-cli/container-config.js
```diff
@@ -0,0 +1,34 @@
+/**
+ * Sparkswap configuration for using the cli inside the container. (Note the rpcCertPath is relative to being inside the container)
+ *
+ * It will use the user defined rpc address, user and password if they exist, otherwise
```
treygriffith: we won't use user-defined rpc address in this file anymore.

docker-compose.yml
```diff
@@ -28,7 +28,6 @@ services:
       - NODE_ENV=production
       - DATA_DIR=${DATA_DIR:-}
       - NETWORK=${NETWORK}
-      - RPC_ADDRESS=${RPC_ADDRESS:-}
```
treygriffith: you should also remove it from the .env file samples as it won't be useful anymore

docker/sparkswapd/Dockerfile
```diff
@@ -39,6 +42,15 @@ COPY --from=builder /home/app/proto /home/app/proto
 
 # Copy node_modules
 COPY --from=builder /home/app/node_modules /home/app/node_modules
+COPY --from=builder /home/app/broker-cli/node_modules /home/app/broker-cli/node_modules
+
+ARG USER_HOME=/root
```
treygriffith: add a comment about why we need this, e.g. "Broker CLI expects config to be at ~/.sparkswap, so we need the default docker user's home directory to install our container config into"

PR #448, opened 2019-03-30
Fix - Add Docker installation steps to install script

bradleystachurski:
## Description
The PR adds better help messages for missing installations of Docker and Docker Compose based on the user's operating system (only Mac and Linux are currently supported). Previously, we were exiting the script and telling the user to "please install it before continuing."

The script also detects the distribution of Linux running the installation script to provide a link to the correct Docker docs.

Before
---
image

After
---
image

## Related PRs

## Todos
- [ ] Tests
- [ ] Documentation
- [x] Link to Trello

### DIFF
--- a/setup-scripts/install.sh
+++ b/setup-scripts/install.sh
@@ -120,6 +120,7 @@ fi
 msg "Installing node.js@8.11" $WHITE
 nvm install 8.11 --latest-npm
 
+# Check if Docker is installed. Fail install if not
 if [ "$(command -v docker)" == "" ]; then
   msg "Docker is not installed." $RED
   if [[ "$OSTYPE" == "linux-gnu" ]]; then
@@ -140,7 +141,7 @@ if [ "$(command -v docker)" == "" ]; then
   exit 1
 fi
 
-# Ensure version of Docker meets minimum requirements. Warn if version is not the latest
+# Ensure version of Docker meets minimum requirements. Fail install if not
 DOCKER_VERSION=$(docker version | grep Version | head -n 1 | grep -oE '[.0-9]*$')
 compare_versions $DOCKER_VERSION $MIN_DOCKER_VERSION
 if [[ $? == 2 ]]; then
@@ -159,6 +160,7 @@ if [[ $? == 2 ]]; then
   exit 1
 fi
 
+# Check if running the latest version of Docker. Warn if not and continue with install
 compare_versions $DOCKER_VERSION $CURRENT_DOCKER_VERSION
 if [[ $? == 2 ]]; then
   msg "Your version of Docker ($DOCKER_VERSION) isn't up to date. It's recommended that you upgrade to version $CURRENT_DOCKER_VERSION or greater." $YELLOW
@@ -174,6 +176,7 @@ if [[ $? == 2 ]]; then
   fi
 fi
 
+# Check if Docker Compose is installed. Fail install if not
 if [ "$(command -v docker-compose)" == "" ]; then
   msg "Docker Compose is not installed." $RED
   if [[ "$OSTYPE" == "linux-gnu" ]]; then
@@ -191,7 +194,7 @@ if [ "$(command -v docker-compose)" == "" ]; then
   exit 1
 fi
 
-# Ensure version of Docker Compose meets minimum requirements. Warn if version is not the latest
+# Ensure version of Docker Compose meets minimum requirements. Fail install if not
 DOCKER_COMPOSE_VERSION=$(docker-compose version | grep 'docker-compose version' | sed 's/,.*//' | grep -oE '[.0-9]*$')
 compare_versions $DOCKER_COMPOSE_VERSION $MIN_DOCKER_COMPOSE_VERSION
 if [[ $? == 2 ]]; then
@@ -208,6 +211,7 @@ if [[ $? == 2 ]]; then
   exit 1
 fi
 
+# Check if running the latest version of Docker Compose. Warn if not and continue with install
 compare_versions $DOCKER_COMPOSE_VERSION $CURRENT_DOCKER_COMPOSE_VERSION
 if [[ $? == 2 ]]; then
   msg "Your version of Docker Compose ($DOCKER_COMPOSE_VERSION) isn't up to date. It's recommended you upgrade to version $CURRENT_DOCKER_COMPOSE_VERSION or greater." $YELLOW
--- a/setup-scripts/install.sh
+++ b/setup-scripts/install.sh
@@ -144,13 +144,34 @@ fi
 DOCKER_VERSION=$(docker version | grep Version | head -n 1 | grep -oE '[.0-9]*$')
 compare_versions $DOCKER_VERSION $MIN_DOCKER_VERSION
 if [[ $? == 2 ]]; then
-  msg "Your version of Docker is older than the minimum required version. Please update to the latest version of Docker." $RED
+  msg "Your version of Docker ($DOCKER_VERSION) is older than the minimum required version. Please upgrade to version $CURRENT_DOCKER_VERSION or greater." $RED
+
+  if [[ "$OSTYPE" == "linux-gnu" ]]; then
+    # Get the version of Linux distribution
+    LINUX_DISTRO=$(cat /etc/*-release | grep -E '^ID=' | sed 's/^.*=//' | tr -d \")
+    msg "You can find steps to upgrade Docker for Linux using the following link." $WHITE
+    msg "https://docs.docker.com/install/linux/docker-ce/$(echo $LINUX_DISTRO)/" $WHITE
+  elif [[ "$OSTYPE" == "darwin"* ]]; then
+    msg "You can find steps to upgrade Docker for Mac using the following link." $WHITE
+    msg "https://docs.docker.com/docker-for-mac/release-notes/" $WHITE
+  fi
+
   exit 1
 fi
 
 compare_versions $DOCKER_VERSION $CURRENT_DOCKER_VERSION
 if [[ $? == 2 ]]; then
-  msg "Your version of Docker isn't up to date. Please update to the latest version." $YELLOW
+  msg "Your version of Docker ($DOCKER_VERSION) isn't up to date. It's recommended that you upgrade to version $CURRENT_DOCKER_VERSION or greater." $YELLOW
+
+  if [[ "$OSTYPE" == "linux-gnu" ]]; then
+    # Get the version of Linux distribution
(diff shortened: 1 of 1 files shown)


bradleystachurski:
[Add Docker installation steps to the install script](https://trello.com/c/Ehlm35N7/1324-add-docker-installation-steps-to-the-install-script)

setup-scripts/install.sh
```diff
@@ -81,12 +81,39 @@ msg "Installing node.js@8.11" $WHITE
...
+    # Get the version of Linux distribution
+    LINUX_DISTRO=$(cat /etc/*-release | grep -E '^ID=' | sed 's/^.*=//' | tr -d \")
+    msg "Looks like you're using Linux." $GREEN
+    msg "To install Docker using the convenience script for Linux, run" $GREEN
+    msg "    $ curl -fsSL https://get.docker.com -o get-docker.sh" $WHITE
+    msg "    $ sudo sh get-docker.sh" $WHITE
+    msg "Alternatively, you can follow Docker's installation steps for Linux (https://docs.docker.com/install/linux/docker-ce/$(echo $LINUX_DISTRO)/)" $GREEN
+  elif [[ "$OSTYPE" == "darwin"* ]]; then
+    msg "Looks like you're using Mac." $GREEN
+    msg "Please use the following link to download and install Docker for Mac Version 2.0.0.3 or greater (this installs Docker Community Edition and Docker Compose)" $GREEN
```
treygriffith: How about "you can use" instead of "please use" ? Some people may want to do `brew`, etc.

setup-scripts/install.sh
```diff
@@ -81,12 +81,39 @@ msg "Installing node.js@8.11" $WHITE
...
-  msg "Docker Compose is not installed. Please install it before continuing" $RED
+  msg "Docker Compose is not installed." $RED
+  if [[ "$OSTYPE" == "linux-gnu" ]]; then
+    msg "To install Docker Compose using Linux, run" $GREEN
+    msg '    $ curl -L "https://github.com/docker/compose/releases/download/1.23.2/docker-compose-$(uname -s)-$(uname -m)" -o /usr/local/bin/docker-compose' $WHITE
+    msg "    $ sudo chmod +x /usr/local/bin/docker-compose" $WHITE
+    msg "Alternatively, you can follow Docker's installation steps for Linux (https://docs.docker.com/compose/install/)" $GREEN
+  elif [[ "" == "" ]]; then
+    msg "Looks like you're using Mac." $GREEN
+    msg "Please use the following link to download and install Docker for Mac Version 2.0.0.3 or greater (this installs Docker Community Edition and Docker Compose)" $GREEN
```
treygriffith: This is a strange case. Clearly they have docker installed, but they don't have docker-compose installed, which means either:

- they are running an old version of Docker for Mac that didn't have docker-compose
- they installed docker not through Docker for Mac (e.g. using `brew`).

Perhaps we should just link to the Docker Compose docs in this case?

setup-scripts/install.sh
```diff
@@ -81,12 +81,39 @@ msg "Installing node.js@8.11" $WHITE
...
+    msg "https://docs.docker.com/docker-for-mac/release-notes/#docker-community-edition-2003-2019-02-15" $WHITE
+  else
+    msg "Please install Docker before continuing" $RED
```
treygriffith: Docker Compose I think you mean

setup-scripts/install.sh
```diff
@@ -81,12 +81,39 @@ msg "Installing node.js@8.11" $WHITE
...
+    msg "Looks like you're using Mac." $GREEN
+    msg "Please use the following link to download and install Docker for Mac Version 2.0.0.3 or greater (this installs Docker Community Edition and Docker Compose)" $GREEN
+    msg "https://docs.docker.com/docker-for-mac/release-notes/#docker-community-edition-2003-2019-02-15" $WHITE
+  else
+    msg "Please install Docker before continuing." $GREEN
+  fi
   exit 1
 fi
 
 if [ "$(command -v docker-compose)" == "" ]; then
```
treygriffith: do we need to check the version here? We've had issues with people using versions that didn't do variable interpolation the correct way.

setup-scripts/install.sh
```diff
@@ -81,12 +81,39 @@ msg "Installing node.js@8.11" $WHITE
 nvm install 8.11 --latest-npm
 
 if [ "$(command -v docker)" == "" ]; then
```
treygriffith: should we have a minimum version here?

setup-scripts/install.sh
bradleystachurski: Yeah, this is a funky scenario. Linking to the Docker Compose docs makes sense.

setup-scripts/install.sh
bradleystachurski: Capturing our in-person conversation: we'll check the version of both `docker` and `docker-compose`. If the version is older than ~6 months, we'll fail the install script and tell the user to update. If the version is the current version, we'll warn the user and continue with the installation process.

setup-scripts/install.sh
bradleystachurski: Refer to above comment.

setup-scripts/install.sh
```diff
@@ -81,15 +121,68 @@ msg "Installing node.js@8.11" $WHITE
...
+DOCKER_VERSION=$(docker version | grep Version | head -n 1 | grep -oE '[.0-9]*$')
+compare_versions $DOCKER_VERSION $MIN_DOCKER_VERSION
+if [[ $? == 2 ]]; then
+  msg "Your version of Docker is older than the minimum required version. Please update to the latest version of Docker." $RED
   exit 1
 fi
 
+compare_versions $DOCKER_VERSION $CURRENT_DOCKER_VERSION
+if [[ $? == 2 ]]; then
+  msg "Your version of Docker isn't up to date. Please update to the latest version." $YELLOW
```
treygriffith: Maybe something like "Your version of Docker ($DOCKER_VERSION) isn't up to date. It's recommended that you upgrade to the latest version: $CURRENT_DOCKER_VERSION."

setup-scripts/install.sh
```diff
@@ -81,15 +121,68 @@ msg "Installing node.js@8.11" $WHITE
...
+    msg "Please install Docker Compose before continuing" $RED
+  fi
   exit 1
 fi
 
+# Ensure version of Docker Compose meets minimum requirements. Warn if version is not the latest
+DOCKER_COMPOSE_VERSION=$(docker-compose version | grep 'docker-compose version' | sed 's/,.*//' | grep -oE '[.0-9]*$')
+compare_versions $DOCKER_COMPOSE_VERSION $MIN_DOCKER_COMPOSE_VERSION
+if [[ $? == 2 ]]; then
+  msg "Your version of Docker Compose is older than the minimum required version. Please update to the latest version of Docker Compose." $RED
```
treygriffith: Just like above, include the actual versions in here. And also include a link to docs or something similar to give them a path to upgrade.

setup-scripts/install.sh
```diff
@@ -81,15 +121,68 @@ msg "Installing node.js@8.11" $WHITE
...
+    msg "Please install Docker before continuing." $GREEN
+  fi
+  exit 1
+fi
+
+# Ensure version of Docker meets minimum requirements. Warn if version is not the latest
+DOCKER_VERSION=$(docker version | grep Version | head -n 1 | grep -oE '[.0-9]*$')
+compare_versions $DOCKER_VERSION $MIN_DOCKER_VERSION
+if [[ $? == 2 ]]; then
+  msg "Your version of Docker is older than the minimum required version. Please update to the latest version of Docker." $RED
```
treygriffith: I would at least link them to the Docker documentation for the recommended version so they have a way to upgrade.

setup-scripts/install.sh
```diff
@@ -25,11 +25,51 @@ NC='\033[0m' # No Color
...
+CURRENT_DOCKER_VERSION=18.09
+CURRENT_DOCKER_COMPOSE_VERSION=1.24
+
 # print a message with a color, like msg "my message" $GRAY
 msg () {
   echo -e "${GRAY}${TAG}${NC}:  $2$1${NC}"
 }
 
+# Compares two versions using dot notation (e.g. 1.2.3 < 1.2.4)
+compare_versions () {
```
treygriffith: why does this return 0, 1, and 2? (I'm not too familiar with bash)

dannypaz: @treygriffith looks like these are return codes. 0 == latest version, 1 == need patch version upgrade but whatever, 2 == you're out of date

setup-scripts/install.sh
bradleystachurski: Good call. I spoke with @dannypaz earlier about how we want to maintain `$CURRENT_DOCKER_VERSION`. We decided to manually update the variable and not include patches as opposed to using `curl` to grab the latest version. Because we're manually updating, I've changed the wording to indicate the user should get `$CURRENT_DOCKER_VERSION` or greater.

setup-scripts/install.sh
bradleystachurski: Yup, it's a bit funky since I believe bash return values are typically used to indicate the status of a command upon exit. An alternative would be assigning the comparison to a variable, then checking the value of the variable for conditional logic.

setup-scripts/install.sh
```diff
@@ -81,15 +121,106 @@ msg "Installing node.js@8.11" $WHITE
...
+    msg "Please install Docker before continuing." $GREEN
+  fi
   exit 1
 fi
 
+# Ensure version of Docker meets minimum requirements. Warn if version is not the latest
+DOCKER_VERSION=$(docker version | grep Version | head -n 1 | grep -oE '[.0-9]*$')
+compare_versions $DOCKER_VERSION $MIN_DOCKER_VERSION
+if [[ $? == 2 ]]; then
+  msg "Your version of Docker ($DOCKER_VERSION) is older than the minimum required version. Please upgrade to version $CURRENT_DOCKER_VERSION or greater." $RED
```
treygriffith: [nit] "or greater" is kind of weird when we use CURRENT_DOCKER_VERSION

setup-scripts/install.sh
```diff
@@ -81,15 +121,106 @@ msg "Installing node.js@8.11" $WHITE
 nvm install 8.11 --latest-npm
 
 if [ "$(command -v docker)" == "" ]; then
```
dannypaz: @bradleystachurski for each section of the install script, we have a comment that is like `# Ensure nvm is installed` or `#install nodejs`, we should make sure we have comments for each of the docker sections.

Example: `# Checking docker installation versions` etc..

dannypaz:
The changes look fine to me. @treygriffith you'll still need to review

PR #449, opened 2019-03-30
Use locked Relayer proto files

treygriffith:
## Description
This change commits the Relayer's proto files to the Broker's version control rather than downloading it from GitHub during the build.

To support local development it still supports downloading from GitHub using the `build-local` command.

Because Docker can't do conditional copying, both the local and packaged proto files are copied into the built container.

## Todos
- [x] Tests
- [x] Documentation

### DIFF
--- a/broker-daemon/relayer/relayer-client.js
+++ b/broker-daemon/relayer/relayer-client.js
@@ -1,5 +1,5 @@
 const path = require('path')
-const { readFileSync } = require('fs')
+const { readFileSync, existsSync } = require('fs')
 const { credentials } = require('grpc')
 const caller = require('grpc-caller')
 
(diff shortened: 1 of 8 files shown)


broker-daemon/relayer/relayer-client.js
```diff
@@ -13,10 +13,19 @@ const consoleLogger = console
 consoleLogger.debug = console.log.bind(console)
 
 /**
+ * Path for Proto files for the Relayer when doing local development
  * @constant
  * @type {string}
  * @default
  */
+const LOCAL_RELAYER_PROTO_PATH = './proto-local/relayer.proto'
```
dannypaz: @treygriffith what was the thought behind the local relayer path? Would we run into issues if we just updated the original `proto/relayer.proto` path?

dannypaz:
[Lock Relayer proto version and distribute with broker](https://trello.com/c/thYp2064/1329-lock-relayer-proto-version-and-distribute-with-broker)

broker-daemon/relayer/relayer-client.js
treygriffith: @dannypaz my goal was to still enable local development but without the risk of overwriting the file in version control. If we pulled into `proto/relayer.proto` it would update in version control and could be inadvertently committed.

PR #454, opened 2019-04-03
Install: Reduce dependencies in install script

treygriffith:
## Description
This change reduces the overall number of dependencies in the install script to make installation easier and have fewer required pre-cursor steps.

Specifically, it removes dependency on:
- `git` (replaced with `curl` and `tar`)
- `npm` (replaced with `bash`)
- `node` and `nvm` (removed installation of the CLI instead, relying on #445 )
- `openssl` (published openssl in purpose-built container)

## Related PRs
#445 was a required pre-cursor

### DIFF
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -6,7 +6,7 @@
 # Options:
 # -c, --no-cli                    do not copy certs for local cli installation
 # -d, --no-docker                 do not build docker images
-# -e=, --external-address=        your public IP address (removes prompt)
+# -e=, --external-address=        your public IP address (used for cert generation)
 # -i, --no-identity               does not generate keys for the daemons identity
 # -n, --no-certs                  does not re-generate tls certs for the daemon
 #
@@ -79,8 +79,7 @@ esac
 done
 
 if [ "$EXTERNAL_ADDRESS" == "" ]; then
-  echo "Please provide your public IP address or hostname"
-  read EXTERNAL_ADDRESS
+  EXTERNAL_ADDRESS="localhost"
 fi
 
 # Create an openssl docker image if the command does not exist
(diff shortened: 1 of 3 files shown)


setup-scripts/install.sh
```diff
@@ -301,12 +294,7 @@ cd broker
 
 # Set up environment
 msg "Setting up your Broker" $WHITE
-npm run env-setup -- -n=$NETWORK
-
-# Install CLI
-msg "Installing the CLI" $WHITE
-npm install -g ./broker-cli
```
dannypaz: Since we removed the cli (because of nvm), we fail to "install" the config on a users machine. This script now throws the error

```
[⚡ Sparkswap Installer]:  Creating random username and password
sed: can't read /root/.sparkswap/config.js: No such file or directory
```

scripts/build.sh
```diff
@@ -142,8 +153,8 @@ if [[ -f "$ID_PRIV_KEY" ]]; then
 elif [[ -f "$ID_PUB_KEY" ]]; then
   echo "WARNING: ID Public Key already exists for Broker Daemon. Skipping ID generation"
 elif [ "$NO_IDENTITY" != "true" ]; then
-  openssl ecparam -name prime256v1 -genkey -noout -out $ID_PRIV_KEY
-  openssl ec -in $ID_PRIV_KEY -pubout -out $ID_PUB_KEY
+  openssl ecparam -name prime256v1 -genkey -noout > $ID_PRIV_KEY
```
dannypaz: does `-noout` affect this? at first glance it looks weird

treygriffith: `do not print the ec parameter`, we want the key only.

PR #455, opened 2019-04-03
add logo-based script branding

treygriffith:
## Description
This changes the ASCII art we display when building Sparkswap to be more consistent with other Sparkswap branding.

Screen Shot 2019-04-03 at 9 47 37 AM

### DIFF
--- a/setup-scripts/install.sh
+++ b/setup-scripts/install.sh
@@ -94,7 +94,20 @@ case $i in
 esac
 done
 
-msg "You're about to install Sparkswap. Good for you!" $GREEN
+echo ""
+echo "  :DMMMMMMM:                                                                   "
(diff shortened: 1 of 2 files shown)


setup-scripts/install.sh
```diff
@@ -94,7 +94,20 @@ case $i in
 esac
 done
 
-msg "You're about to install Sparkswap. Good for you!" $GREEN
+echo ""
```
dannypaz: by doing this, the logo will show up twice (cause we'll run the build spec). Is that OK?

dannypaz: Actually, will show up 3 times for the current install

dannypaz: You can ignore this. Its behind a build flag

PR #456, opened 2019-04-03
Fix - Install script exits when `compare_versions` is called

bradleystachurski:
## Description
This PR fixes the implementation of `compare_verison`, which was recently introduced to the install script.

`compare_versions` previously returned a value based on the result of the string comparison (comparing versions using dot notation i.e. "1.2.3" < "1.2.4"). Since the install script uses `set -e`, the script fails when any command return a non 0 exit code. `compare_versions` has been updated to set a variable to the result of the comparison.

## Related PRs

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Link to Trello

### DIFF
--- a/setup-scripts/install.sh
+++ b/setup-scripts/install.sh
@@ -32,6 +32,9 @@ MIN_DOCKER_COMPOSE_VERSION=1.23.0
 CURRENT_DOCKER_VERSION=18.09
 CURRENT_DOCKER_COMPOSE_VERSION=1.24
 
+# Used for generating download URL
+RECOMMENDED_DOCKER_COMPOSE_VERSION=1.24.0
+
 # print a message with a color, like msg "my message" $GRAY
 msg () {
   echo -e "${GRAY}${TAG}${NC}:  $2$1${NC}"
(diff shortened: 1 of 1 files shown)


bradleystachurski:
[Install Script exits when `compare_versions` is called](https://trello.com/c/9b5XAcEp/1375-install-script-exits-when-compareversions-is-called)

dannypaz:
@bradleystachurski A few comments

```
[⚡ Sparkswap Installer]:      $ curl -fsSL https://get.docker.com -o get-docker.sh
[⚡ Sparkswap Installer]:      $ sudo sh get-docker.sh
```

We should remove the `$` on this example and we should also consolidate the commands we want the user to run... as an example

```
curl -fsSL https://get.docker.com -o get-docker.sh && sudo sh get-docker.sh
```

It'll be much easier for the user to copy-pasta. I would add this for both commands

dannypaz:
@bradleystachurski from our talk earlier, if the user is on linux and they dont have docker installed, we can assume they also dont have docker-compose and should prompt them for both instead of failing twice

PR #457, opened 2019-04-03
Restart LND on errors

treygriffith:
## Description
This change causes LND to restart on errors. LND will still be locked and will need to be manually unlocked, but this is better than simply being down.

In particular, this solves an edge case where LND starts up faster than the underlying litecoind/bitcoind and shuts down. Instead, it will restart in a locked state.


dannypaz:
@treygriffith LGTM

PR #458, opened 2019-04-03
Fix - Install script for non root Linux users

bradleystachurski:
## Description
This PR adds support for running the install script as a user on Linux instead of as root. The main change is adding the user to the `docker` group after installation.

## Related PRs

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Link to Trello


bradleystachurski:
[Bug: Unable to install docker-compose for non-root user in install script](https://trello.com/c/Scfc9GpM/1386-bug-unable-to-install-docker-compose-for-non-root-user-in-install-script)

PR #459, opened 2019-04-04
remove cli installation from setup script

martineehrlich:
## Description
We no longer do this by default. The cli is installed in the dockerfile, and if users want to use the remote version these instructions are in the docs. 

## Related PRs

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Link to Trello


dannypaz:
@martineehrlich this is no longer needed as the CLI has been removed from the install script in https://github.com/sparkswap/broker/pull/454

PR #460, opened 2019-04-04
Feature - Stacked Orderbook

bradleystachurski:
## Description
This PR changes the orderbook UI.

Previously the ordrebook was represented as a table of bids side-by-side to a table of asks. The new UI represents the orderbook as asks stacked above bids.

This change was made since the typical representation of an orderbook in traditional finance along with most crypto exchanges represent the orderbook as stacked. In traditional finance, this is often referred to as the Depth of Market (DOM).

Previous Orderbook
---
image

New Orderbook
---
image

## Related PRs

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Link to Trello


bradleystachurski:
[Orderbook should be stacked](https://trello.com/c/WdU45oQV/1362-orderbook-should-be-stacked)

dannypaz:
@bradleystachurski this looks awesome. @treygriffith can you review too

broker-cli/commands/orderbook.js
```diff
@@ -17,37 +17,117 @@ const EVENT_TYPES = Object.freeze({
...
- * @param {Array<Object>} bids - an object with price and depth
+ * @param {Array<Object>} asks - an array of objects with price and depth
+ * @param {Array<Object>} bids - an array of object with price and depth
  */
 function createUI (market, asks, bids) {
   console.clear()
+  const windowHeight = size.get().height
+
+  // Fill as many orders as the screen allows less other info displayed
+  const maxLengthPerSide = Math.floor((windowHeight - 12) / 2)
```
treygriffith: might constantize `12` here to make it more descriptive and easier to edit

broker-cli/commands/orderbook.js
```diff
@@ -17,37 +17,117 @@ const EVENT_TYPES = Object.freeze({
...
+  const gapTable = new Table({
+    head: [],
+    rowHeights: [1],
+    // Only need bottom border, so remove all other characters
     chars: {
       'top': '',
       'top-mid': '',
       'top-left': '',
       'top-right': '',
-      'bottom': '',
```
treygriffith: worthwhile to specify the character you want? I know it's in there by default, but might clarify this section

broker-cli/commands/orderbook.js
```diff
@@ -150,6 +278,7 @@ async function orderbook (args, opts, logger) {
     })
 
     process.stdout.on('resize', function () {
+      console.clear()
```
treygriffith: this seems redundant since you re-clear at the top of `createUI`, right?

broker-cli/commands/orderbook.js
```diff
@@ -162,19 +291,51 @@ async function orderbook (args, opts, logger) {
...
+  }
 
-function calculateTableWidths (windowWidth) {
-  const borderOffset = 4
-  const numTables = 2
-  const mainTableWidth = Math.round((windowWidth - borderOffset) / numTables)
-  const innerTableWidth = Math.round((mainTableWidth - borderOffset) / numTables)
+  const needsGrayColoring = lastZeroIndex > -1
+
+  const addColoring = (text) => {
```
treygriffith: why not define this as a function outside of `formatText` ?

broker-cli/commands/orderbook.js
```diff
@@ -162,19 +291,51 @@ async function orderbook (args, opts, logger) {
...
- * @returns {Object} with mainTableWidth and innerTableWidth
+ * Takes a number formatted as a string and applies coloring so all digits following
+ * the last significant digit are grayed out
+ * @param {string} text
+ * @param {string} type - represents whether to format as a bid, ask, or depth
+ * @returns {string}
  */
+function formatText (text, type) {
+  let lastZeroIndex = -1
+  for (let i = 0; i < text.length; i++) {
```
treygriffith: you might pull this into its own function, like `findLastZero`. Also consider using a function like: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/lastIndexOf which may be helpful if you reverse the string first, for example.

bradleystachurski: Pulled this into another function.

I think if we reverse the string then look for the index of the first occurrence of a non-`0`, then do a `lastIndexOf` `0` and find the unreversed index, it might be less clear what's going on with roughly the same length function. You okay keeping it as-is in a separate function?

broker-cli/commands/orderbook.js
```diff
@@ -58,10 +150,73 @@ function createUI (market, asks, bids) {
...
+      'padding-left': 0,
+      'padding-right': 0
+    }
+  })
+
+  gapTable.push([' ', ' '])
+
+  const displayAsks = (asks) => {
+    // Asks are added to the table with highest (worst) price at top, so need to sort
+    [...asks].sort((a, b) => {
```
treygriffith: You can just do `asks.slice().reverse()` here instead. Might be a little simpler and probably faster.

broker-cli/commands/orderbook.js
```diff
@@ -58,10 +150,73 @@ function createUI (market, asks, bids) {
...
+    colWidths: [COLUMN_WIDTHS, COLUMN_WIDTHS],
+    style: {
+      'padding-left': 0,
+      'padding-right': 0
+    }
+  })
+
+  gapTable.push([' ', ' '])
+
+  const displayAsks = (asks) => {
```
treygriffith: does this need to be defined inside this block or can it be defined outside?

broker-cli/commands/orderbook.js
```diff
@@ -58,10 +150,73 @@ function createUI (market, asks, bids) {
...
+    [...asks].sort((a, b) => {
+      return b.price - a.price
+    }).forEach((ask) => {
+      const formattedPrice = formatText(ask.price, FORMAT_TYPES.ASK)
+      const formattedDepth = formatText(ask.amount, FORMAT_TYPES.DEPTH)
+      asksTable.push([{ hAlign: 'center', content: formattedPrice }, { hAlign: 'center', content: formattedDepth }])
+    })
+  }
+
+  if (asks.length <= maxLengthPerSide) {
```
treygriffith: You should try to unify the logic between asks and bids. Per our conversation, perhaps one way to do this is to treat them identically (e.g. pushing spaces, x more..., etc on the end of the array) until the very end, when you reverse the asks.

PR #463, opened 2019-04-05
Whitelist HTTP Gateway methods

treygriffith:
## Description
This change:
- fixes some bugs with the grpcGateway
- adds a new configuration for a whitelist for HTTP methods

The intention behind both of these is to allow a Broker to be deployed with an HTTP proxy in place without opening up all methods for usage over HTTP. This may be useful for developing a read-only viewer, for example.

This is pretty much a bandaid on a more fully-featured solution which would include real authentication for the HTTP solution and would have multiple levels of authentication by method.

## Todos
- [x] Tests
- [x] Documentation

### DIFF
--- a/broker-daemon/bin/sparkswapd.js
+++ b/broker-daemon/bin/sparkswapd.js
@@ -49,6 +49,7 @@ program
   // Broker RPC configuration options
   .option('--rpc.address [rpc-address]', 'Add a host/port to listen for daemon RPC connections', validations.isHost, config.rpc.address)
   .option('--rpc.http-proxy-address [rpc-http-proxy-address]', 'Add a host/port to listen for HTTP RPC proxy connections', validations.isHost, config.rpc.httpProxyAddress)
+  .option('--rpc.http-proxy-methods [rpc-http-proxy-methods]', 'Comma-separated methods for the HTTP RPC Proxy', program.String, config.rpc.httpProxyMethods)
   .option('--rpc.user [rpc-user]', 'Broker rpc user name', program.String, config.rpc.user)
   .option('--rpc.pass [rpc-pass]', 'Broker rpc password', program.String, config.rpc.pass)
   .option('--rpc.pub-key-path [rpc-pub-key-path]', 'Location of the public key for the broker\'s rpc', validations.isFormattedPath, config.rpc.pubKeyPath)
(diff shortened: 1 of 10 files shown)


broker-daemon/bin/sparkswapd.js
```diff
@@ -49,6 +49,7 @@ program
   // Broker RPC configuration options
   .option('--rpc.address [rpc-address]', 'Add a host/port to listen for daemon RPC connections', validations.isHost, config.rpc.address)
   .option('--rpc.http-proxy-address [rpc-http-proxy-address]', 'Add a host/port to listen for HTTP RPC proxy connections', validations.isHost, config.rpc.httpProxyAddress)
+  .option('--rpc.http-proxy-methods [rpc-http-proxy-methods]', 'Comma-separated methods for the HTTP RPC Proxy', program.String, config.rpc.httpProxyMethods)
```
dannypaz: @treygriffith we need to add an ENV for this in `start-sparkswapd.sh` if we want users to be able to modify it

broker-daemon/utils/create-http-server.js
```diff
@@ -33,7 +33,7 @@ function handle404 (logger, req, res) {
  * @param {string} opts.logger
  * @returns {ExpressApp}
  */
-function createHttpServer (protoPath, rpcAddress, { disableAuth = false, enableCors = false, privKeyPath, pubKeyPath, logger }) {
+function createHttpServer (protoPath, rpcAddress, { disableAuth = false, enableCors = false, privKeyPath, pubKeyPath, httpMethods = ['*'], logger }) {
```
dannypaz: @treygriffith should we just default `httpMethods` at the point of use? We default the value in multiple places but then have a `WILDCARD` var in `grpc-gateway`

broker-daemon/utils/grpc-gateway.js
```diff
@@ -71,7 +80,10 @@ const middleware = (protoFiles, grpcLocation, credentials = grpc.credentials.cre
         s.methods.forEach(m => {
           if (m.options[GRPC_API_OPTION_ID]) {
             supportedMethods.forEach(httpMethod => {
-              if (m.options[GRPC_API_OPTION_ID][httpMethod]) {
+              // We should limit methods to those included on the whitelist
+              const shouldGateway = whitelist.includes(WILDCARD) || whitelist.includes(m.options[GRPC_API_OPTION_ID][httpMethod])
```
dannypaz: [nit] `shouldGateway` is a tough name because we are in the gateway... unfortunately I dont have suggestions on how we can make variable describe the action better. Maybe something like `allowMethod` or `permitMethod`

dannypaz:
@treygriffith change is good. I tested this w/ basic auth and works +1

PR #468, opened 2019-04-06
Version bump/0.5.0 beta rc2

treygriffith:


treygriffith:
Wrong version, we're on 0.6

PR #472, opened 2019-04-06
Configure self-signed certs

treygriffith:
## Description
When we use self-signed certificates with the Broker we need to configure the internal proxy differently than if they are issued by e.g. LetsEncrypt.

This adds a new configuration for self-signed certificates to make either work.

## Related PRs
#470 

## Todos
- [x] Tests


.circleci/config.yml
```diff
@@ -15,16 +15,16 @@ jobs:
       - checkout
 
       # Download and cache dependencies
-      - restore_cache:
-          keys:
-          - v3-dependencies-{{ checksum "package.json" }}
+      # - restore_cache:
```
dannypaz: @treygriffith probably didnt mean to check this in

treygriffith: I did in fact mean to check this in because circle cache was failing. I'm considering removing the cache altogether. Talk me off the edge.

dannypaz: @treygriffith lol we have had all kinds of issue with the caching. Looks like the best we can do right now is:

1. make sure npm is updated (6.9.x)
2. remove the lock and push the change
3. add the lock file back and push the change

PR #475, opened 2019-04-07
FIX dont initialize blockorders before engines are validated

dannypaz:
## Description
This PR solves a bug where, on restart, the broker tries to re-work indeterminate orders (orders that are still executing/placed/create) however, if the engines are in a non-validated state, the orders will fail and subsequently be cancelled.

We now check if engines have been validated before we try to work indeterminate orders.

## Related PRs

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Link to Trello

### DIFF
--- a/broker-daemon/index.js
+++ b/broker-daemon/index.js
@@ -183,16 +183,17 @@ class BrokerDaemon {
    */
   async initialize () {
     try {
-      await this.initializeMarkets(this.marketNames)
-
       // Starts the validation of all engines on the broker. We do not await this
       // function because we want the validations to run in the background as it
       // can take time for the engines to be ready
       const enginesAreValidated = this.validateEngines()
 
-      // We run this asynchronously so that it doesn't block the start of the rpc
-      // server. This function will not return until the engines are validated
-      this.initializeBlockOrder(enginesAreValidated)
+      // Since these are potentially long-running operations, we run them in parallel to speed
+      // up BrokerDaemon startup time.
+      await Promise.all([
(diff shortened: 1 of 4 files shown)


broker-daemon/index.js
```diff
@@ -186,18 +197,11 @@ class BrokerDaemon {
       // Since these are potentially long-running operations, we run them in parallel to speed
       // up BrokerDaemon startup time.
       await Promise.all([
+        this.validateEngines(),
```
treygriffith: I don't think we want to wait for the engines to be ready to start the broker

dannypaz:
[Wait until engines are ready before resolving indeterminate Block Orders](https://trello.com/c/IfkhIvEH/1401-wait-until-engines-are-ready-before-resolving-indeterminate-block-orders)

broker-daemon/index.js
```diff
@@ -183,20 +183,16 @@ class BrokerDaemon {
...
-      // the validation will be retried on the engine until successful or until final failure.
-      this.validateEngines()
+      // Starts the validation of all engines on the broker. We do not await this
+      // function because we want the validations to run in the background as it
+      // can take time for the engines to be ready
+      const enginesAreValidated = this.validateEngines()
+
+      // We run this asynchronously so that it doesn't block the start of the rpc
+      // server. This function will not return until the engines are validated
+      this.initializeBlockOrder(enginesAreValidated)
```
treygriffith: Any way we can await only the first part of this? What happens if the RPC starts but the block orders are not yet indexed?

PR #481, opened 2019-04-08
FIX - update proto to build for other languages

dannypaz:
## Description
This PR modifies out broker.proto file to conform to protocol buffer spec so that users can build the broker proto file for other languages.

**Changes**
1. changed scoping of `RelayerStatuses`, `OrderbookStatuses` and `Engine Statuses`
2. changed `OK` of `RelayerStatuses` to `RELAYER_OK` to avoid conflicts with `OrderbookStatuses`

When running the original broker.proto file through protoc, we received the following errors:

```
http.proto: File not found.
google/api/annotations.proto: Import "http.proto" was not found or had errors.
google/api/annotations.proto:30:3: "HttpRule" is not defined.
broker.proto: Import "google/api/annotations.proto" was not found or had errors.
broker.proto:99:3: "UNAVAILABLE" is already defined in "broker.rpc".
broker.proto:99:3: Note that enum values use C++ scoping rules, meaning that enum values are siblings of their type, not children of it.  Therefore, "UNAVAILABLE" must be unique within "broker.rpc", not just within "RelayerStatuses".
broker.proto:107:3: "OK" is already defined in "broker.rpc".
broker.proto:107:3: Note that enum values use C++ scoping rules, meaning that enum values are siblings of their type, not children of it.  Therefore, "OK" must be unique within "broker.rpc", not just within "OrderbookStatuses".
broker.proto:109:3: "NOT_SYNCED" is already defined in "broker.rpc".
broker.proto:109:3: Note that enum values use C++ scoping rules, meaning that enum values are siblings of their type, not children of it.  Therefore, "NOT_SYNCED" must be unique within "broker.rpc", not just within "OrderbookStatuses".
```

**Protoc Usage:**
`protoc -I broker-daemon/proto -I broker-daemon/proto/google/api broker-daemon/proto/broker.proto --go_out=plugins=grpc:broker-daemon/proto`

## Related PRs

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Link to Trello

### DIFF
--- a/broker-cli/proto/broker.proto
+++ b/broker-cli/proto/broker.proto
@@ -69,50 +69,27 @@ enum Side {
   ASK = 1;
 }
 
-/**
- * Statuses of an engine
- */
-enum EngineStatuses {
-  // Default state of the engine
-  UNKNOWN = 0;
-  // Engine is unavailable
-  UNAVAILABLE = 1;
-  // Wallet does not exist
-  NEEDS_WALLET = 2;
-  // Wallet exists but is not locked
-  LOCKED = 3;
-  // Wallet is unlocked but the configuration of the node and chain sync status are not known
-  UNLOCKED = 4;
-  // Engine is not yet synced to chain
-  NOT_SYNCED = 5;
-  // Engine is validated and ready for use
-  VALIDATED = 6;
-}
-
-/**
- * Statuses of the relayer
- */
-enum RelayerStatuses {
-  // Relayer is available
-  OK = 0;
-  // Relayer is not available
-  UNAVAILABLE = 1;
-}
-
-/**
- * Statuses of an orderbook
- */
-enum OrderbookStatuses {
-  // Orderbook is synced
-  OK = 0;
-  // Orderbook is not synced
-  NOT_SYNCED = 1;
-}
 
 /**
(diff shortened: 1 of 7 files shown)


dannypaz:
[Broker Proto not compatible with protobuf](https://trello.com/c/68zDs0ZF/1410-broker-proto-not-compatible-with-protobuf)

broker-cli/proto/broker.proto
```diff
@@ -69,50 +69,27 @@ enum Side {
...
+    // Orderbook is synced
+    OK = 0;
+    // Orderbook is not synced
+    NOT_SYNCED = 1;
+  }
+
+  // Statuses of the relayer
+  enum RelayerStatuses {
+    // Relayer is available
+    RELAYER_OK = 0;
```
treygriffith: Maybe we should prefix all of them? e.g. `RELAYER_UNAVAILABLE` and `ORDERBOOK_OK` ?

dannypaz:
@treygriffith I added a task for us to update circleci to check w/ `protoc`. we'll have to look into it

broker-cli/proto/broker.proto
dannypaz: This makes sense since they are status codes that we are setting. ill make the change

PR #482, opened 2019-04-08
remove node dependency

martineehrlich:
## Description
We want to remove node as a dependency from the build script completely. We can do this by using bash instead of node to find the package.json version.

## Related PRs

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Link to Trello


martineehrlich:
[Remove node dependency from build script](https://trello.com/c/UihyApEM/1413-remove-node-dependency-from-build-script)

PR #485, opened 2019-04-10
Feature - Wallet Summary

dannypaz:
Requires <https://github.com/sparkswap/lnd-engine/pull/198> before merging

## Description
This PR creates a new command `sparkswap wallet summary <symbol>` on the CLI (and Broker). This endpoint provides on-chain fee and transaction info about the users wallet. The primary purpose of this feature is to allow the user to have information related to accounting for their node 

Screen Shot 2019-04-15 at 2 45 38 PM

This PR does not include paging for transactions

## Related PRs

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Link to Trello


dannypaz:
[Record funds spent by the broker opening channels](https://trello.com/c/g86thBm8/1332-record-funds-spent-by-the-broker-opening-channels)

broker-daemon/proto/broker.proto
```diff
@@ -1874,4 +1908,12 @@ service WalletService {
       body: "*"
     };
   }
+
+  /**
+  */
+  rpc WalletSummary (WalletSummaryRequest) returns (WalletSummaryResponse) {
```
treygriffith: Add an example of usage

broker-cli/commands/wallet.js
```diff
@@ -531,6 +532,46 @@ async function unlock (args, opts, logger) {
...
+ */
+async function history (args, opts, logger) {
+  const { symbol } = args
+  const { rpcAddress } = opts
+
+  try {
+    const client = new BrokerDaemonClient(rpcAddress)
+    const { transactions = [] } = await client.walletService.walletHistory({ symbol })
+
+    if (!transactions) {
```
treygriffith: since you initialize `transactions`, this code path will only ever be run if `walletHistory` returns a value like `false` or `0` for transactions.

broker-cli/commands/wallet.js
```diff
@@ -531,6 +532,46 @@ async function unlock (args, opts, logger) {
...
+      logger.info(`No summary available for ${symbol}`)
```
treygriffith: This would almost be better in the table if the transactions length was zero so that you could have visual confirmation that there are not transactions (vs thinking it is a bug)

broker-cli/commands/wallet.js
```diff
@@ -531,6 +532,46 @@ async function unlock (args, opts, logger) {
...
+      head: ['Type', 'Amount', 'Fees', 'Timestamp', 'Transaction', 'Height', 'Pending'],
+      style: { head: ['gray'] }
+    })
+
+    transactions.forEach(({ type, amount, fees, timestamp, transactionHash, blockHeight, pending }) => {
+      transactionTable.push([type, amount, fees, timestamp, transactionHash, blockHeight, pending])
+    })
+
+    logger.info('')
+    logger.info('  Transactions:')
```
treygriffith: is this necessary?

broker-daemon/broker-rpc/wallet-service/wallet-history.js
```diff
@@ -0,0 +1,30 @@
...
+
+  if (!engine) {
+    logger.error(`Could not find engine: ${symbol}`)
+    throw new Error(`No engine found for symbol: ${symbol}`)
+  }
+
+  const transactions = await engine.getChainTransactions()
+
+  if (!transactions.length) {
+    logger.debug('No transactions found in walletHistory')
```
treygriffith: [nit] might this be better as `${transactions.length} transactions found in walletHistory`?

broker-cli/commands/wallet.js
dannypaz: I just want to check the length here. Thanks for catching this

broker-cli/commands/wallet.js
dannypaz: nah, not necessary

PR #487, opened 2019-04-11
Use latest npm to avoid cache issues

dannypaz:
## Description
Update our CI scripts to have npm at the latest version when we run specs or install things from cache. We are hoping this will prevent caching issues from occurring during `master` builds of the broker

## Related PRs

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Link to Trello

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -14,6 +14,9 @@ jobs:
     steps:
       - checkout
 
+      # Make sure we are on the latest npm version
+      - run: npm i -g npm
+
       # Download and cache dependencies
       - restore_cache:
           keys:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -14,6 +14,9 @@ jobs:
     steps:
       - checkout
 
+      # Make sure we are on the latest npm version
+      - run: npm i -g npm
```
treygriffith: Should we not just make sure we're on a specific version of npm rather than the generic latest?

dannypaz: @treygriffith ill stick it to a major

PR #489, opened 2019-04-12
Trade History

martineehrlich:
## Description
We want a command to get us history of a users trades. In order to do this, we need to get all orders and fills from the respective stores and filter based on their statuses (completed, executing for orders and accepted, executed for fills).

We also wanted to add dates that each status for an order/fill has been reached, so I added a state machine plugin to document this information when persisting the record.

## Related PRs

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Link to Trello

### DIFF
--- a/broker-cli/commands/order/index.js
+++ b/broker-cli/commands/order/index.js
@@ -16,13 +16,15 @@ const SUPPORTED_COMMANDS = Object.freeze({
   STATUS: 'status',
   CANCEL: 'cancel',
   SUMMARY: 'summary',
-  CANCEL_ALL: 'cancel-all'
+  CANCEL_ALL: 'cancel-all',
+  TRADE_HISTORY: 'trade-history'
 })
 
 const status = require('./status')
 const cancel = require('./cancel')
 const summary = require('./summary')
 const cancelAll = require('./cancel-all')
+const tradeHistory = require('./trade-history')
 
 module.exports = (program) => {
   program
@@ -58,6 +60,9 @@ module.exports = (program) => {
         case SUPPORTED_COMMANDS.CANCEL_ALL:
           opts.market = validations.isMarketName(market)
           return cancelAll(args, opts, logger)
+
+        case SUPPORTED_COMMANDS.TRADE_HISTORY:
+          return tradeHistory(args, opts, logger)
       }
     })
     .command(`order ${SUPPORTED_COMMANDS.SUMMARY}`, 'View your orders')
@@ -73,5 +78,7 @@ module.exports = (program) => {
     .option('--rpc-address [rpc-address]', RPC_ADDRESS_HELP_STRING, validations.isHost)
     .command(`order ${SUPPORTED_COMMANDS.CANCEL_ALL}`, 'Cancel all block orders on market')
     .option('--market <marketName>', MARKET_NAME_HELP_STRING, validations.isMarketName, null, true)
+    .option('--rpc-address [rpc-address]', RPC_ADDRESS_HELP_STRING, validations.isHost)
+    .command(`order ${SUPPORTED_COMMANDS.TRADE_HISTORY}`, 'Cancel all block orders on market')
     .option('--rpc-address [rpc-address]', RPC_ADDRESS_HELP_STRING, validations.isHost)
 }
--- a/broker-cli/commands/order/index.js
+++ b/broker-cli/commands/order/index.js
@@ -79,6 +79,6 @@ module.exports = (program) => {
     .command(`order ${SUPPORTED_COMMANDS.CANCEL_ALL}`, 'Cancel all block orders on market')
     .option('--market <marketName>', MARKET_NAME_HELP_STRING, validations.isMarketName, null, true)
     .option('--rpc-address [rpc-address]', RPC_ADDRESS_HELP_STRING, validations.isHost)
-    .command(`order ${SUPPORTED_COMMANDS.TRADE_HISTORY}`, 'Cancel all block orders on market')
+    .command(`order ${SUPPORTED_COMMANDS.TRADE_HISTORY}`, 'Get information about completed and processing trades')
     .option('--rpc-address [rpc-address]', RPC_ADDRESS_HELP_STRING, validations.isHost)
 }
--- a/broker-cli/commands/order/trade-history.js
+++ b/broker-cli/commands/order/trade-history.js
@@ -16,7 +16,7 @@ async function tradeHistory (args, opts, logger) {
   try {
     const client = new BrokerDaemonClient(rpcAddress)
     const tradeHistory = await client.orderService.getTradeHistory({})
-    logger.info(tradeHistory)
+    logger.info(JSON.stringify(tradeHistory, null, 2))
   } catch (e) {
     logger.error(handleError(e))
   }
--- a/broker-cli/commands/order/trade-history.js
+++ b/broker-cli/commands/order/trade-history.js
@@ -2,12 +2,11 @@ const BrokerDaemonClient = require('../../broker-daemon-client')
 const { handleError } = require('../../utils')
 
 /**
- * sparkswap order status
+ * sparkswap order trade-history
  *
- * ex: `sparkswap order status Aar_w9XuTtUqeqeaac5liIMR-Lqf1dJfKZikTkhJ'
+ * ex: `sparkswap order trade-history'
  *
  * @param {Object} args
- * @param {string} args.blockOrderId
  * @param {Object} opts
  * @param {string} opts.rpcaddress
  * @param {Logger} logger
@@ -16,8 +15,8 @@ async function tradeHistory (args, opts, logger) {
   const { rpcAddress } = opts
   try {
     const client = new BrokerDaemonClient(rpcAddress)
-    const blockOrderResult = await client.orderService.getTradeHistory({})
-    logger.info(blockOrderResult)
+    const tradeHistory = await client.orderService.getTradeHistory({})
+    logger.info(tradeHistory)
   } catch (e) {
     logger.error(handleError(e))
   }
--- a/broker-cli/commands/order/trade-history.js
+++ b/broker-cli/commands/order/trade-history.js
@@ -1,0 +1,26 @@
+const BrokerDaemonClient = require('../../broker-daemon-client')
+const { handleError } = require('../../utils')
+
+/**
+ * sparkswap order status
+ *
+ * ex: `sparkswap order status Aar_w9XuTtUqeqeaac5liIMR-Lqf1dJfKZikTkhJ'
+ *
+ * @param {Object} args
+ * @param {string} args.blockOrderId
+ * @param {Object} opts
+ * @param {string} opts.rpcaddress
+ * @param {Logger} logger
+ */
+async function tradeHistory (args, opts, logger) {
+  const { rpcAddress } = opts
+  try {
+    const client = new BrokerDaemonClient(rpcAddress)
+    const blockOrderResult = await client.orderService.getTradeHistory({})
+    logger.info(blockOrderResult)
+  } catch (e) {
+    logger.error(handleError(e))
+  }
+};
+
+module.exports = tradeHistory
--- a/broker-cli/proto/broker.proto
+++ b/broker-cli/proto/broker.proto
@@ -419,7 +419,7 @@ message Fill {
     FILLED = 1;
     // The swap has been executed on the Payment Channel Networks.
     EXECUTED = 2;
-    // The preimage has been returned to the Relayer by the Maker.
+    // The fill has encountered a failure.
     REJECTED = 4;
     // The fill has been cancelled.
     CANCELLED = 5;
@@ -435,7 +435,7 @@ message Fill {
     string filled = 2;
     // The date the fill has encountered a failure.
     string executed = 3;
-    // The date the preimage has been returned to the Relayer by the Maker.
+    // The date the fill has encountered a failure.
     string rejected = 5;
     // The date the fill has been cancelled.
     string cancelled = 6;
--- a/broker-cli/proto/broker.proto
+++ b/broker-cli/proto/broker.proto
@@ -395,6 +395,8 @@ message Order {
   string side = 10;
   // Dates that the order has reached each status
   Date dates = 11;
+  // Amount of the order filled, expressed in a decimal string of common units for a currency (e.g. bitcoins for BTC). This can be any amount greater than zero and less than or equal to the amount in the order.
+  string fill_amount = 12;
 }
 
 /**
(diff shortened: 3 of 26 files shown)


broker-cli/proto/broker.proto
```diff
@@ -538,6 +538,47 @@ message GetBlockOrdersRequest {
...
+message TradeOrder {
+  string order_id = 1;
+  string block_order_id = 2;
+  string base_symbol = 3;
+  string counter_symbol = 4;
+  string side = 5;
+  string base_amount = 6;
+  string counter_amount = 7;
+  string state = 8;
+  string date_created = 9;
```
treygriffith: These dates might be better as a submessage so that they can be managed separately.

broker-cli/proto/broker.proto
```diff
@@ -538,6 +538,47 @@ message GetBlockOrdersRequest {
...
 
+message TradeOrder {
+  string order_id = 1;
+  string block_order_id = 2;
+  string base_symbol = 3;
+  string counter_symbol = 4;
+  string side = 5;
+  string base_amount = 6;
+  string counter_amount = 7;
+  string state = 8;
```
treygriffith: do we call this `state` other places, or `status`?

broker-cli/proto/broker.proto
```diff
@@ -538,6 +538,47 @@ message GetBlockOrdersRequest {
   string market = 1;
 }
 
+message TradeOrder {
```
treygriffith: can `TradeOrder` and `TradeFill` be consolidated?

broker-cli/commands/order/index.js
```diff
@@ -74,4 +79,6 @@ module.exports = (program) => {
     .command(`order ${SUPPORTED_COMMANDS.CANCEL_ALL}`, 'Cancel all block orders on market')
     .option('--market <marketName>', MARKET_NAME_HELP_STRING, validations.isMarketName, null, true)
     .option('--rpc-address [rpc-address]', RPC_ADDRESS_HELP_STRING, validations.isHost)
+    .command(`order ${SUPPORTED_COMMANDS.TRADE_HISTORY}`, 'Cancel all block orders on market')
```
treygriffith: bad description

broker-daemon/proto/broker.proto
```diff
@@ -360,16 +360,43 @@ message Order {
     CANCELLED = 6;
   }
 
+  message Date {
```
treygriffith: I'm pretty sure this name will conflict with the one for Fills. And in any event, it's not very descriptive. Perhaps `OrderDates` ? or `OrderStateDates`?

broker-daemon/proto/broker.proto
```diff
@@ -360,16 +360,43 @@ message Order {
...
+    string date_rejected = 5;
+    // The date the order has been cancelled by the user.
+    string date_cancelled = 6;
+  }
+
   // Unique ID assigned by the Relayer for this order.
   string order_id = 1;
   // Status of the order on the Relayer.
-  OrderStatus order_status = 2;
+  OrderStatus status = 2;
```
treygriffith: why make this change?

broker-daemon/proto/broker.proto
```diff
@@ -360,16 +360,43 @@ message Order {
...
+  string block_order_id = 7;
+  // The base currency for the order
+  string base_symbol = 8;
+  // The counter currency for the order
+  string counter_symbol = 9;
+  // The side of the market that the order is taking.
+  string side = 10;
+  // Dates that the order has reached each status
+  Date dates = 11;
+  // Amount of the order filled, expressed in a decimal string of common units for a currency (e.g. bitcoins for BTC). This can be any amount greater than zero and less than or equal to the amount in the order.
```
treygriffith: I would note that this will be an empty string if the order has not been filled.

broker-cli/commands/order/trade-history.js
```diff
@@ -0,0 +1,25 @@
...
+ * @param {Object} opts
+ * @param {string} opts.rpcaddress
+ * @param {Logger} logger
+ */
+async function tradeHistory (args, opts, logger) {
+  const { rpcAddress } = opts
+  try {
+    const client = new BrokerDaemonClient(rpcAddress)
+    const tradeHistory = await client.orderService.getTradeHistory({})
+    logger.info(JSON.stringify(tradeHistory, null, 2))
```
treygriffith: I assume there is a ticket for turning this into a table?

broker-cli/proto/broker.proto
```diff
@@ -794,6 +854,117 @@ service OrderService {
...
+   * ```
+   *
+   * > The above command will retrieve the trades on the Broker and will print:
+   *
+   * ```javascript
+   * {
+   *   "orders": [
+   *     {
+   *       "orderId": "Ky13eTmuhZY-NIbqIng_in3D",
+   *       "orderStatus": "PLACED",
```
treygriffith: placed orders shouldn't show up in trade history, right?

broker-cli/proto/broker.proto
```diff
@@ -794,6 +854,117 @@ service OrderService {
...
+   *         "dateCreated": "2019-04-12T18:22:42.519Z",
+   *         "datePlaced": "2019-04-12T18:22:42.534Z",
+   *         "dateExecuting": "",
+   *         "dateCompleted": "",
+   *         "dateRejected": "",
+   *         "dateCancelled": ""
+   *       }
+   *     }
+   *   ],
+   *   "fills": []
```
treygriffith: fills would be good to add to the docs too since they have a different data structure

broker-daemon/block-order-worker/index.js
```diff
@@ -847,6 +847,38 @@ class BlockOrderWorker extends EventEmitter {
     })
   }
 
+  /**
+   * Get all completed trade info
+   * @returns {Object} result
+   * @returns {Array<Object>} result.orders all orders for a user
+   * @returns {Array<Object>} result.fills all fills for a user
+   */
+  async getTrades () {
```
treygriffith: It's weird to name this `getTrades` if it retrieves all orders and fills, not just wants that are "trades" (i.e. executed). Why not filter inside of this function?

broker-daemon/broker-rpc/order-service/get-trade-history.js
```diff
@@ -0,0 +1,29 @@
...
+ * @param {Object} responses
+ * @param {function} responses.GetTradeHistoryResponse - constructor for GetTradeHistoryResponse messages
+ * @return {responses.GetTradeHistoryResponse}
+ */
+
+async function getTradeHistory ({ logger, blockOrderWorker }, { GetTradeHistoryResponse }) {
+  try {
+    const { orders, fills } = await blockOrderWorker.getTrades()
+
+    const filteredOrders = orders.filter(order => order.status === 'COMPLETED' || order.status === 'EXECUTING')
```
treygriffith: See my other comment, but I think this filtering would be better in the block order worker (and should use constants for status)

broker-daemon/state-machines/fill-state-machine.js
```diff
@@ -140,7 +156,7 @@ const FillStateMachine = StateMachine.factory({
    * @returns {Object} Data to attach to the state machine
    */
   data: function ({ store, logger, relayer, engines }) {
-    return { store, logger, relayer, engines, fill: {} }
+    return { store, logger, relayer, engines, fill: {}, dates: {} }
```
treygriffith: would be nice if this could be initialized elsewhere rather than by the implementing state machine.

broker-daemon/state-machines/plugins/state-machine-dates.js
```diff
@@ -0,0 +1,36 @@
...
+   * @returns {Object} Key value of observers
+   */
+  get observers () {
+    const plugin = this
+
+    return {
+      onEnterState: function (lifecycle) {
+        if (!plugin.skipTransitions.includes(lifecycle.transition)) {
+          // set the date the state was entered, for example, 'dateCreated'
+          this.dates[`date${this.state.charAt(0).toUpperCase()}${this.state.slice(1)}`] = new Date()
```
treygriffith: since this is already in a property called `dates`, can't we just call this `created`?

e.g. this.dates['created'] = blah

That seems clear enough to me and avoids any string manipulation

broker-cli/commands/order/trade-history.js
martineehrlich: Just added one. I felt like this PR was already huge but ill do it once this is merged.

broker-cli/proto/broker.proto
martineehrlich: Oh oops, this is leftover from testing.

broker-daemon/block-order-worker/index.js
martineehrlich: Good call, I'll make that change.

broker-daemon/state-machines/plugins/state-machine-dates.js
martineehrlich: Yeah i think you're right. I had this before we had that dates object.

broker-daemon/state-machines/fill-state-machine.js
martineehrlich: Do you mean by the plugin?

broker-cli/commands/order/trade-history.js
treygriffith: @martineehrlich it actually might be something to punt for later. It's a nice to have.

broker-daemon/proto/broker.proto
```diff
@@ -538,6 +589,13 @@ message GetBlockOrdersRequest {
   string market = 1;
 }
 
+message GetTradeHistoryResponse {
```
treygriffith: Add a description of this message

broker-daemon/proto/broker.proto
```diff
@@ -388,15 +417,27 @@ message Fill {
...
   }
+
+  message FillStateDates {
+    // The date the fill has been created on the Relayer.
+    string created = 1;
+    // The date the fill has been accepted as the fill for the given order.
+    string filled = 2;
+    // The date the swap has been executed on the Payment Channel Networks.
+    string executed = 3;
+    // The date the preimage has been returned to the Relayer by the Maker.
```
treygriffith: This description is incorrect

broker-daemon/proto/broker.proto
```diff
@@ -388,15 +417,27 @@ message Fill {
...
     EXECUTED = 2;
     // The preimage has been returned to the Relayer by the Maker.
-    COMPLETED = 3;
-    // The fill has encountered a failure.
     REJECTED = 4;
     // The fill has been cancelled.
     CANCELLED = 5;
   }
+
+  message FillStateDates {
```
treygriffith: Add a description for this

broker-daemon/proto/broker.proto
```diff
@@ -360,6 +360,21 @@ message Order {
     CANCELLED = 6;
   }
 
+  message OrderStateDates {
```
treygriffith: Add a description for this message

broker-daemon/block-order-worker/index.spec.js
```diff
@@ -2438,4 +2440,107 @@ describe('BlockOrderWorker', () => {
       expect(res).to.eql([firstBlockOrder])
     })
   })
+
+  describe('getTrades', () => {
```
treygriffith: probably need a test for serialization

broker-daemon/models/fill.js
```diff
@@ -398,6 +399,41 @@ class Fill {
...
+    const {
+      fill,
+      state,
+      error,
+      dates
+    } = fillObject
+
+    const { order } = fill
+
+    const baseAmountFactor = CONFIG.currencies.find(({ symbol }) => symbol === order.baseSymbol).quantumsPerCommon
```
treygriffith: Do we use the engine config for this in other places? I think we are trying to transition to using engine config where possible.

broker-daemon/state-machines/order-state-machine.js
```diff
@@ -42,6 +43,9 @@ const OrderStateMachine = StateMachine.factory({
     new StateMachineLogging({
       skipTransitions: [ 'goto' ]
     }),
+    new StateMachineDates({
```
treygriffith: I'm assuming that this plugin needs to be added before `StateMachinePersistence` in order to make sure that the date of the latest transition is added before being saved to the db.

If that's the case, it may be worth testing that behavior, or at least documenting the requirement.

broker-daemon/models/fill.js
martineehrlich: Can you link an example of where we use engineConfig?

treygriffith: https://github.com/sparkswap/broker/blob/master/broker-daemon/broker-rpc/wallet-service/commit.js#L51

broker-daemon/state-machines/order-state-machine.js
martineehrlich: Hm, I haven't found it to be an issue. All the dates get saved properly 🤔 but I'll do some digging.

broker-daemon/models/fill.js
treygriffith: Maybe we should pass in this kind of config when we initialize the model rather than pulling it out during serialization?

broker-daemon/state-machines/plugins/state-machine-dates.spec.js
```diff
@@ -0,0 +1,67 @@
+const { expect, timekeeper } = require('test/test-helper')
+
+const StateMachineDates = require('./state-machine-dates')
+const StateMachine = require('../state-machine')
+
+describe('StateMachineLogging', () => {
```
treygriffith: bad name

broker-daemon/block-order-worker/index.js
dannypaz: @martineehrlich ^^

broker-daemon/state-machines/fill-state-machine.js
```diff
@@ -406,6 +424,30 @@ const FillStateMachine = StateMachine.factory({
...
+  const {
+    fill,
+    state,
+    error,
+    dates
+  } = fillObject
+
+  const serializedFill = fill.serialize()
+
+  Object.assign(serializedFill, {
```
treygriffith: [nit] generally you don't want to `Object.assign` to an object that you didn't create as you may unintentionally mutate something. Two ways to do this that would avoid that issue:

```
Object.assign({}, serializedFill, { other stuff })
```

or

```
return {
  fillStatus: state.toUpperCase(),
  error: error ? error.toString() : undefined,
  dates,
  ...serializedFill
}
```

PR #491, opened 2019-04-12
Feature: add created date to order summary table

treygriffith:
## Description
Displays the date that the Block Order was created in the Order Summary CLI command.

## Todos
- [x] Tests
- [x] Documentation

### DIFF
--- a/broker-cli/commands/order/summary.js
+++ b/broker-cli/commands/order/summary.js
@@ -15,8 +15,8 @@ function createUI (market, orders) {
   const unitWidth = Math.floor(windowWidth / 16)
 
   const orderTable = new Table({
-    head: ['Order ID', 'Side', 'Amount', 'Limit Price', 'Time', 'Status'],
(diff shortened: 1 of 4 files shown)


broker-cli/commands/order/summary.js
```diff
@@ -15,8 +15,8 @@ function createUI (market, orders) {
   const unitWidth = Math.floor(windowWidth / 16)
 
   const orderTable = new Table({
-    head: ['Order ID', 'Side', 'Amount', 'Limit Price', 'Time', 'Status'],
-    colWidths: [5 * unitWidth, unitWidth, 3 * unitWidth, 3 * unitWidth, unitWidth, 2 * unitWidth],
+    head: ['Order ID', 'Status', 'Side', 'Amount', 'Limit Price', 'Time', 'Created At'],
+    colWidths: [2 * unitWidth, unitWidth, unitWidth, 3 * unitWidth, 3 * unitWidth, unitWidth, 3 * unitWidth],
```
dannypaz: @treygriffith as an aside (the PR looks good), have we tried removing colWidths? I think that would force caporal to handle the column widths dynamically so we didn't need to calculate anything

PR #492, opened 2019-04-13
Feature - Lexicographic ordering for base64 encoded order IDs

bradleystachurski:
## Description
This PR adds a utility function `encodeBase64Lex` which converts a base64 encoded string to an ASCII representation that preserves lexicographic ordering.

This is useful since order IDs are prepended with the UNIX timestamp they were created, so using an encoding scheme that preserves lexicographic ordering allows the user to view orders sorted by date.

## Related PRs

## Todos
- [x] Tests
- [x] Documentation
- [x] Link to Trello

### DIFF
--- a/broker-daemon/utils/base64-lex.spec.js
+++ b/broker-daemon/utils/base64-lex.spec.js
@@ -20,8 +20,10 @@ describe('encodeBase64Lex', () => {
   it('strips off base64 padding', () => {
     // base64 encoding will have padding of only 1 or 2 '='
     // (3 bytes are represented in 4 6-bit base64 digits, i.e. encoding is mod 3)
-    expect(encodeBase64Lex('12345=')).to.be.eql('pqrst')
-    expect(encodeBase64Lex('12345==')).to.be.eql('pqrst')
+    const onePad = Buffer.from('ffff', 'hex') // base64 encoded is '//8='
+    const twoPad = Buffer.from('ff', 'hex') // base64 encoded is '/w=='
+    expect(encodeBase64Lex(onePad)).to.be.eql('zzw')
+    expect(encodeBase64Lex(twoPad)).to.be.eql('zk')
   })
 
   it('preserves lexicographic ordering for numbers vs uppercase letters', () => {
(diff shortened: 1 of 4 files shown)


bradleystachurski:
[Block order ids are not in order](https://trello.com/c/HEZf4uOG/1212-block-order-ids-are-not-in-order)

broker-daemon/utils/base64-lex.spec.js
```diff
@@ -0,0 +1,110 @@
...
+
+    const expected = [
+      { timestamp: 1554505000000, lexId: 'M9US9-' },
+      { timestamp: 1554831515000, lexId: 'M9nNak' },
+      { timestamp: 1554914185000, lexId: 'M9sQXF' }
+    ]
+
+    // Convert timestamps to base64 encoding then convert to lex ordering
+    const lexIds = []
+    timestamps.forEach(ts => {
```
treygriffith: why not just map timestamps to an array of `lexIds`? Then you can just do `.sort()` without a custom sort function.

broker-daemon/utils/generate-id.js
```diff
@@ -20,19 +21,7 @@ function generateId () {
...
- * Convert a string from Base64 to [Base64url]{@link https://tools.ietf.org/html/rfc4648}
- * @param  {string} base64Str - Base64 encoded String
- * @returns {string}           Base64url encoded string
- */
-function urlEncode (base64Str) {
-  return base64Str
-    .replace(/=/g, '')
-    .replace(/\+/g, '-')
-    .replace(/\//g, '_')
+  return base64Lex(id.toString('base64'))
```
treygriffith: should `base64Lex` take a buffer instead of a base64 string? You can do conversion to base64 inside of your function.

broker-daemon/utils/base64-lex.spec.js
```diff
@@ -0,0 +1,102 @@
...
+    expect(encodeBase64Lex('12345=')).to.be.eql('pqrst')
+    expect(encodeBase64Lex('12345==')).to.be.eql('pqrst')
+  })
+
+  it('preserves lexicographic ordering for numbers vs uppercase letters', () => {
+    base64LowerIndex = 'A' // base64 index for 'A' is 0
+    base64HigherIndex = '0' // base64 index for '0' is 52
+
+    // ASCII character codes would incorrectly sort based on base64 for this case
+    expect(base64LowerIndex.charCodeAt(0)).to.be.greaterThan(base64HigherIndex.charCodeAt(0))
```
treygriffith: I think this test would be better as:
1. two pieces of data, probably represented as Buffers or hex strings
2. use regular base64 and show they are misordered (`.toString('base64')`)
3. use the lex base64 and show they are correctly ordered.

Right now it exercises the logic, but not the code as much.

PR #494, opened 2019-04-17
remove docker override info

dannypaz:
## Description
This PR removes references to `docker-compose.override` because we no longer uses a setup that leverages that file. When we specify `COMPOSE_FILE` in the `.env` file, this will actually override any defaults from `docker-compose` making the comment about the override file obsolete.

## Related PRs

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Link to Trello


dannypaz:
[Docker override cleanup](https://trello.com/c/d22eq14n/1477-docker-override-cleanup)

PR #509, opened 2019-05-01
Fix: Relayer Health should show UNAVAILABLE when it is

treygriffith:
## Description
The recent change in specific statuses for the Relayer vs Orderbook left a bug in returning health status to the Broker Client, where the Relayer being down would be interpreted as `OK` because the constant was actually `undefined`.

~~We should consider switching which value is the default (currently `OK`), but that would be a pretty confusing breaking change for users of the API.~~ **Edit**: Done in this PR.

## Related PRs
Introduced in #481 

## Todos
- [x] Tests


dannypaz:
Looks good to me

dannypaz:
[Deadline exceeded for Relayer results in "OK" status](https://trello.com/c/iEY5nKCK/1527-deadline-exceeded-for-relayer-results-in-ok-status)

PR #511, opened 2019-05-02
Fix - Cache relayer healthcheck

bradleystachurski:
## Description
This PR adds a caching utility that wraps a function call and caches results for a defined period.

This is useful for caching the results of `healthCheck`. When the relayer loses connection with a broker, the broker will check if the relayer is available by calling `healthCheck` for every single active order. This needlessly spams the relayer with requests.

## Related PRs

## Todos
- [x] Tests
- [x] Documentation
- [x] Link to Trello


bradleystachurski:
[Relayer gets spammed with healthchecks on a connection break](https://trello.com/c/1YRHpQ8k/1462-relayer-gets-spammed-with-healthchecks-on-a-connection-break)

broker-daemon/utils/cached-call.js
```diff
@@ -0,0 +1,53 @@
...
+ */
+class CachedCall {
+  /**
+   * Create a new CachedCall instance.
+   * @param {Function} promiseFn - function that returns a promise
+   * @param {number} ttl - milliseconds for previous calls to remain valid
+   */
+  constructor (promiseFn, ttl = TIME_TO_LIVE) {
+    this.promiseFn = promiseFn
+    this.lastCallTime = 0
```
martineehrlich: 0 should be in a const

broker-daemon/utils/cached-call.js
```diff
@@ -0,0 +1,53 @@
...
+    this.lastCallPromise = undefined
```
martineehrlich: I feel like this needs some commentary, what is lastCallPromise?

dannypaz: Setting this to null should have the same effect instead of defining a value as undefined XD

broker-daemon/utils/cached-call.js
```diff
@@ -0,0 +1,53 @@
+/**
+ * Default time for the cached call to live. Calls made after this time will
+ * make another call to the promise function.
+ * @constant
+ * @type {number}
+ * @default
+ */
+const TIME_TO_LIVE = 5000
```
dannypaz: [nit] call this CACHE_TTL instead (its weird to see it spelled out)

broker-daemon/block-order-worker/index.js
```diff
@@ -892,13 +895,12 @@ class BlockOrderWorker extends EventEmitter {
...
    */
-  async relayerIsAvailable () {
-    try {
-      await this.relayer.adminService.healthCheck({})
-      return true
-    } catch (e) {
-      return false
+  relayerIsAvailable () {
+    if (!this.cachedCall) {
+      this.cachedCall = new CachedCall(() => this.relayer.adminService.healthCheck({}))
```
dannypaz: @bradleystachurski suggestion for how to handle `CachedCall` without needing an anonymous function.

Have you tried having 2 arguments to CachedCall:

```
func = this.relayer.adminService.healthCheck
params = {}
new CachedCall(func, params)
```

then inside of the CachedCall class

```
return this.lastCallFn(this.lastCallParams)
```

bradleystachurski: Capturing our in-person chat. Going to keep the anonymous function here since the fat arrow keeps the context of `this` without being overly verbose with `Function.prototype.bind`.

broker-daemon/utils/cached-call.js
```diff
@@ -0,0 +1,63 @@
+/**
+ * Default time for the cached call to live. Calls made after this time will
+ * make another call to the promise function.
```
dannypaz: would make a note that this is in `ms`

PR #525, opened 2019-05-24
Feature/healthcheck json

dannypaz:
## Description
This PR adds a `--json` flag to the healthcheck command to allow the healthcheck to be parsable by other programs

## Related PRs

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Link to Trello

### DIFF
--- a/broker-cli/commands/health-check.js
+++ b/broker-cli/commands/health-check.js
@@ -69,7 +69,7 @@ async function healthCheck (args, opts, logger) {
     const res = await client.adminService.healthCheck({})
 
     if (json) {
-      return console.log(JSON.stringify(res))
+      return logger.info(JSON.stringify(res, null, 2))
     }
 
     const {
--- a/broker-cli/commands/health-check.js
+++ b/broker-cli/commands/health-check.js
@@ -58,16 +58,25 @@ const ENGINE_STATUS_CODES = Object.freeze({
  */
 
 async function healthCheck (args, opts, logger) {
-  const { rpcAddress } = opts
+  const {
+    rpcAddress,
+    json
+  } = opts
 
   try {
     const client = new BrokerDaemonClient(rpcAddress)
+
+    const res = await client.adminService.healthCheck({})
+
+    if (json) {
+      return console.log(JSON.stringify(res))
+    }
 
     const {
       engineStatus = [],
       orderbookStatus: orderbookStatuses = [],
       relayerStatus = STATUS_CODES.UNKNOWN
-    } = await client.adminService.healthCheck({})
+    } = res
 
     const healthcheckTable = new Table({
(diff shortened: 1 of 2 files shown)


dannypaz:
[Relayer not starting / being down does not alert us via PagerDuty](https://trello.com/c/NL03isAq/1589-relayer-not-starting-being-down-does-not-alert-us-via-pagerduty)

broker-cli/commands/health-check.js
```diff
@@ -58,16 +58,25 @@ const ENGINE_STATUS_CODES = Object.freeze({
...
+    json
+  } = opts
 
   try {
     const client = new BrokerDaemonClient(rpcAddress)
 
+    const res = await client.adminService.healthCheck({})
+
+    if (json) {
+      return console.log(JSON.stringify(res))
```
bradleystachurski: Not sure about this: do we need to use `logger` for this to show up in scalyr?

Also, low priority, in `wallet.js` we do a pretty print (`JSON.stringify(transactions, null, 2`). Should we create a ticket to remove the pretty printing so it can also be consumed by other programs more easily?

dannypaz: @bradleystachurski this is such a good observation. Thanks for the catch.

1. we should use logger, i will make the change
2. pretty printing is actually what we want to do, i've added this to the code

Pretty printing is only there to see it visually (the same for transactions). If there are any use-cases where this doesn't allow someone to use the `--json` flag effectively, then we can modify.

For us, our current use would be to make sure that different keys in the response are present and I wanted to do this through JSON instead of having to parse the stdout table (although they are arguably the same)

As an example, from the way the code is now, we can see if relayer has RELAYER_OK by running the following:

```
sparkswap healthcheck --json | grep 'RELAYER_OK' | wc -l
```

Returning 1 if we are OK and returning 0 if something is wrong

broker-cli/commands/health-check.spec.js
```diff
@@ -57,15 +64,21 @@ describe('healthCheck', () => {
...
   it('makes a request to the broker', async () => {
     await healthCheck(args, opts, logger)
     expect(healthCheckStub).to.have.been.called()
   })
 
+  it('returns json if the user specifies a json flag', async () => {
+    opts.json = true
+    await healthCheck(args, opts, logger)
+    expect(instanceTableStub.push).to.not.have.been.called()
+    expect(jsonStub.stringify).to.have.been.calledWith(healthCheckResponse)
```
treygriffith: also test that it logs the output of the stringify?

broker-cli/commands/health-check.js
bradleystachurski: Gotcha, that makes sense.

PR #527, opened 2019-05-28
Fix - Calculate Active Funds performance issues

dannypaz:
## Description
This PR fixes our previous approach to calculating active funds (iterating over each blockorder) to instead grabbing all records for fills/orders and then filtering by market or side.

The issue is that once a broker has greater than 20k orders, the performance degrades to where an order would take > 20 seconds to finish placing due to the calculating of funds.

This change is targeted to temporarily fix performance issues

**Changes**
1. New methods to get fills/orders for a range of block order ids
2. New methods to calculate active funds for a given set of orders/fills
3. modified `calculateActiveFunds`
4. added `market` to fills and orders

## Related PRs

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Link to Trello


dannypaz:
[Broker performance issues when placing orders](https://trello.com/c/ZOMv88MG/1609-broker-performance-issues-when-placing-orders)

broker-daemon/block-order-worker/index.js
```diff
@@ -310,28 +310,33 @@ class BlockOrderWorker extends EventEmitter {
...
+    // If we have no block orders that are available for our particular side, then
+    // we assume no active orders are available and can simply return 0
+    if (!blockOrders.length) {
+      return { activeOutboundAmount: Big(0), activeInboundAmount: Big(0) }
     }
 
-    return { activeOutboundAmount, activeInboundAmount }
+    // We grab the firstId and lastId to create a blockorder range so that we can
+    // grab all fills and orders to calculate active funds
+    const firstId = blockOrders[blockOrders.length - 1].id
```
treygriffith: Are the block orders guaranteed to be sorted by ID?

broker-daemon/block-order-worker/index.js
```diff
@@ -310,28 +310,33 @@ class BlockOrderWorker extends EventEmitter {
...
+    const lastId = blockOrders[0].id
+
+    const orders = await BlockOrder.getOrdersForRange(this.ordersStore, firstId, lastId)
```
treygriffith: Couldn't this include orders that belong to block orders that have their ID in the range, but are not actually active? If we are filtering out the inactive ones anyway, why even filter by the block order id range? Is it a performance optimization?

broker-daemon/block-order-worker/index.js
```diff
@@ -462,7 +468,11 @@ class BlockOrderWorker extends EventEmitter {
     }
 
     const allRecords = await getRecords(this.store, BlockOrder.fromStorage.bind(BlockOrder), queryOptions)
-    const recordsForMarket = allRecords.filter((record) => record.marketName === market)
+    let filteredRecords = allRecords.filter((record) => record.marketName === market)
+
+    if (options.side) {
+      filteredRecords = filteredRecords.filter((r) => r.side === options.side)
```
treygriffith: doesn't having a `limit` option conflict with filtering after the fact? You end up limiting before filtering which seems counterintuitive.

broker-daemon/models/order.js
```diff
@@ -470,6 +470,20 @@ class Order {
...
   }
+
+  /**
+   * Create a set of options that can be passed to a Level `createReadStream` call
+   * that limits the set to orders that belong to the given range of blockOrderIds.
+   * @param {string} firstId - blockOrderId of the start of our block order range
+   * @param {string} lastId - blockOrderId of the end of our block order range
+   * @returns {Object} Options object that can be used in {@link https://github.com/Level/levelup#createReadStream}
+   */
+  static rangeForIds (firstId, lastId) {
```
treygriffith: don't we have this defined somewhere else?

broker-daemon/models/fill.js
```diff
@@ -425,7 +425,7 @@ class Fill {
    * Create a set of options that can be passed to a LevelUP `createReadStream` call
    * that limits the set to fills that belong to the given blockOrderId.
    * This works because all fills are prefixed with their blockOrderId and the Delimiter.
-   * @param {string} blockOrderId - of of the block order to create a range for
+   * @param {string} blockOrderId - id of the block order to create a range for
    * @returns {Object} Options object that can be used in {@link https://github.com/Level/levelup#createReadStream}
    */
   static rangeForBlockOrder (blockOrderId) {
```
treygriffith: You may want to implement this by using the `rangeForIds` and pass the same id twice

broker-daemon/models/block-order.js
```diff
@@ -476,6 +476,100 @@ class BlockOrder {
...
+        const { order, state, error, dates } = JSON.parse(value)
+        return { order: Order.fromObject(key, order), state, error, dates }
+      },
+      // limit the orders we retrieve to those that belong to this blockOrder, i.e. those that are in
+      // its prefix range.
+      Order.rangeForIds(first, last)
+    )
+  }
+
+  static async getFillsForRange (store, first, last) {
```
treygriffith: Seems odd to have this defined as a static on the BlockOrder. Any reason why you did that?

broker-daemon/models/block-order.js
```diff
@@ -476,6 +476,100 @@ class BlockOrder {
...
+    )
+  }
+
+  /**
+   * @param {Array<Order>} orders
+   * @returns {Object} res
+   * @returns {Big} res.inbound
+   * @returns {Big} res.outbound
+   */
+  static async activeAmountsForOrders (orders) {
```
treygriffith: since this is so duplicative of `BlockOrder#activeOutboundAmount`, is there a way to refactor this or that method to use the same code paths?

broker-daemon/block-order-worker/index.js
bradleystachurski: Block order IDs should be sorted by time so long as there weren't any orders that went through an order lifecycle prior to the lexicographical ordering update.

broker-daemon/block-order-worker/index.js
dannypaz: Yeah, for the performance optimization.

**Original Method:**
1. Grab all block orders
2. For EACH block order, populate orders and fills
3. For EACH block order, Filter for active orders and active fills (happened in the model)
4. For EACH block order, get the active inbound balance and active outbound balance

**New Approach**
1. Grab all block orders
2. Grab all orders/fills for a range of block orders
3. Filter all of the orders/fills to be active
4. Grab the active amounts from each group (order or fill)

The thing that was really killing us was these queries separately on each block order

broker-daemon/block-order-worker/index.js
dannypaz: Yeah, you can either limit records, or filter them, you're not able to do both effectively

dannypaz: This is an issue from the previous PR

broker-daemon/models/block-order.js
dannypaz: @treygriffith I didnt have a strong reason to keep them on the BlockOrder. I only kept the methods here because of `populateOrders` and `populateFills`.

I debated moving this to Fill/Order respectively

broker-daemon/models/block-order.js
dannypaz: Although it seems duplicative, its for different models. I felt like making this more abstract would make it harder to read

dannypaz: I tried doing something like

```
activeAmountsForModel (data, allowedStates)
```

and then trying to make the logic for `if EXECUTING` look nicer. I wasn't happy with it

dannypaz: Moving these to the Fill/Order, respectively, might be a better solution just cause they are probably weird to have on the BlockOrder itself

broker-daemon/models/order.js
dannypaz: we have `rangeForBlockOrder` above this code but it doesn't allow for a range. If we want to consolidate these methods, we could

dannypaz: just saw your comment here: https://github.com/sparkswap/broker/pull/527/files#r288694181 this is what I will do

broker-daemon/block-order-worker/index.js
treygriffith: I guess what I mean is: does `getBlockOrders` make an API guarantee that block orders will be streamed in order? Maybe it does, but we should document that at least if we are relying on it.

broker-daemon/block-order-worker/index.js
treygriffith: Couldn't we just skip to step 2 in the new approach and just grab all orders and fills and filter by active? Are we expecting significant benefit from filtering by block orders first?

dannypaz: Yeah, at this point in the broker we can, if we have another market, then we will have to add an index or change how this section of the code works.

You make a good point here, one that I was trying to think through. By grabbing all of the block orders, we had an assumption that "This range of block orders is only for this type of market", but the current implementation does not take that into account

broker-daemon/block-order-worker/index.js
dannypaz: Ive added a ticket to adjust this, since it is counter intuitive for a limit

broker-daemon/models/fill.js
dannypaz: Ive made this change in a separate PR <https://github.com/sparkswap/broker/pull/529>

broker-daemon/models/order.js
dannypaz: https://github.com/sparkswap/broker/pull/529

broker-daemon/models/block-order.js
treygriffith: I was thinking that `activeInboundAmount` and `activeOutboundAmount` would depend on these methods.

dannypaz: That works for me

broker-daemon/block-order-worker/index.js
dannypaz: No longer need to worry about this because I removed the call to blockorders

broker-daemon/block-order-worker/index.js
```diff
@@ -305,33 +305,33 @@ class BlockOrderWorker extends EventEmitter {
...
-        ])
-        return bo
-      })
-    )
+  async calculateActiveFunds (side) {
+    let orders = await Order.getAllOrders(this.ordersStore)
+    let fills = await Fill.getAllFills(this.fillsStore)
+
+    // Filter out records for a particular side if it exists. This is needed to be able
+    // to calculate inbound/amount amounts for the network status tab.
```
bradleystachurski: [nit] Think you mean `inbound/outbound`

broker-daemon/models/block-order.js
```diff
@@ -476,6 +438,67 @@ class BlockOrder {
...
+      CREATED,
+      FILLED
+    } = FillStateMachine.STATES
+    const response = {
+      inbound: Big(0),
+      outbound: Big(0)
+    }
+    return fills.reduce((acc, { fill }) => {
+      // If the order is not in an active state , then we can simply skip it
+      if (![CREATED, FILLED].includes(fill.state)) {
```
bradleystachurski: General question: is there a reason the `Fill` state is structured differently than the `Order` state? (`orders` have state on same level as `order` where `fills` have state as a property of `fill`)

broker-daemon/models/block-order.js
```diff
@@ -476,6 +438,67 @@ class BlockOrder {
...
+      inbound: Big(0),
+      outbound: Big(0)
+    }
+    return orders.reduce((acc, { order, state }) => {
+      // If the order is not in an active state , then we can simply skip it
+      if (![CREATED, PLACED, EXECUTING].includes(state)) {
+        return acc
+      }
+
+      if (state === OrderStateMachine.STATES.EXECUTING) {
```
bradleystachurski: [nit] `EXECUTING` is destructured above, so should be able to use the more concise definition.

broker-daemon/models/fill.js
```diff
@@ -435,6 +438,21 @@ class Fill {
       lte: `${endId}${DELIMITER}${UPPER_BOUND}`
     }
   }
+
+  /**
+   * Grabs all orders for a range of block orders
```
treygriffith: [nit] this is a bad copypasta

broker-daemon/models/order.js
```diff
@@ -471,6 +474,21 @@ class Order {
       lte: `${endId}${DELIMITER}${UPPER_BOUND}`
     }
   }
+
+  /**
+   * Grabs all orders for a range of block orders
```
treygriffith: [nit] bad copypasta

broker-daemon/models/order.js
```diff
@@ -471,6 +474,21 @@ class Order {
...
+   * @param {sublevel} store
+   * @returns {void}
```
treygriffith: [nit] missing return value

broker-daemon/models/block-order.js
```diff
@@ -476,6 +438,67 @@ class BlockOrder {
...
     return BlockOrder.fromStorage(blockOrderId, value)
   }
+
+  /**
+   * @param {Array<Order>} orders
+   * @returns {Object} res
+   * @returns {Big} res.inbound
+   * @returns {Big} res.outbound
+   */
+  static async activeAmountsForOrders (orders) {
```
treygriffith: maybe this belongs somewhere else, like `BlockOrderWorker`, or `OrderStateMachine`. It doesn't have much to do with the `BlockOrder`.

broker-daemon/models/block-order.js
dannypaz: @bradleystachurski there should be no difference. I think the property exists on both, however this should be the same as the order example.

e.g. `{ fill, state }` in the reduce function and we check with `.includes(state)`

broker-daemon/block-order-worker/index.js
```diff
@@ -241,8 +241,8 @@ class BlockOrderWorker extends EventEmitter {
    * @throws {Error} If there are insufficient outbound or inbound funds
    */
   async checkFundsAreSufficient (blockOrder) {
-    const { marketName, side, outboundSymbol, inboundSymbol } = blockOrder
-    const { activeOutboundAmount, activeInboundAmount } = await this.calculateActiveFunds(marketName, side)
+    const { outboundSymbol, inboundSymbol } = blockOrder
+    const { activeOutboundAmount, activeInboundAmount } = await this.calculateActiveFunds()
```
treygriffith: seems like we should still be passing the side here, right?

broker-daemon/block-order-worker/index.js
```diff
@@ -302,36 +302,93 @@ class BlockOrderWorker extends EventEmitter {
...
-    )
+  async calculateActiveFunds (side) {
+    let orders = await Order.getAllOrders(this.ordersStore)
+    let fills = await Fill.getAllFills(this.fillsStore)
+
+    // Filter out records for a particular side if it exists. This is needed to be able
+    // to calculate inbound/outbound amounts for the network status tab.
+    //
+    // TODO: Allow calculateActiveFunds to use market by indexes on orders/fills
+    if (side) {
```
treygriffith: is it impossible to filter based on market for fills and orders? (is that why the TODO is here?)

broker-daemon/block-order-worker/index.js
dannypaz: yeah we need the side here

broker-daemon/block-order-worker/index.js
dannypaz: nope. we could create a custom method for it, but yeah, that is why I left the TODO. Most of the tradeoffs I made were for performance. if we filter all of those records, it just increases the time for the calls.

PR #528, opened 2019-05-28
Feature - Post Only order type

bradleystachurski:
## Description
This PR adds a new order type, Post Only orders.

With this order type, an order is moved to a cancelled state if any portion of the order would immediately execute when placed. This is a useful feature for market makers that wish to place orders that provide liquidity instead of taking.

**Changes**
1. Update `workLimitBlockOrder` to handle both `GTC` and `PO` orders
2. Add Post Only to `TimeInForce` enum in proto

## Related PRs

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Link to Trello

### DIFF
--- a/broker-cli/proto/broker.proto
+++ b/broker-cli/proto/broker.proto
@@ -310,9 +310,9 @@ service AdminService {
 enum TimeInForce {
   GTC = 0; // Good-til-cancelled, an order that remains in force until cancelled by the user
   // FOK is not supported
-  FOK = 1; // Fill-or-kill, an order that can be completely filled or will not be executed
+  // FOK = 1; // Fill-or-kill, an order that can be completely filled or will not be executed
   // IOC is not supported
-  IOC = 2; // Immediate-or-cancel, an order that will have as much filled immediately as possible with the remainder cancelled
+  // IOC = 2; // Immediate-or-cancel, an order that will have as much filled immediately as possible with the remainder cancelled
   PO = 3; // Post Only, an order that will be rejected if any part of the order would match and execute immediately
 }
 
--- a/broker-cli/proto/broker.proto
+++ b/broker-cli/proto/broker.proto
@@ -309,8 +309,8 @@ service AdminService {
  */
 enum TimeInForce {
   GTC = 0; // Good-til-cancelled, an order that remains in force until cancelled by the user
-  // FOK = 1; // Not supported - Fill-or-kill, an order that can be completely filled or will not be executed
-  // IOC = 2; // Not supported - Immediate-or-cancel, an order that will have as much filled immediately as possible with the remainder cancelled
+  FOK = 1; // Fill-or-kill, an order that can be completely filled or will not be executed. Note: this time restriction is not currently supported.
+  IOC = 2; // Immediate-or-cancel, an order that will have as much filled immediately as possible with the remainder cancelled. Note: this time restriction is not currently supported.
   PO = 3; // Post Only, an order that will be rejected if any part of the order would match and execute immediately
 }
 
--- a/broker-cli/proto/broker.proto
+++ b/broker-cli/proto/broker.proto
@@ -309,8 +309,11 @@ service AdminService {
  */
 enum TimeInForce {
   GTC = 0; // Good-til-cancelled, an order that remains in force until cancelled by the user
+  // FOK is not supported
   FOK = 1; // Fill-or-kill, an order that can be completely filled or will not be executed
+  // IOC is not supported
   IOC = 2; // Immediate-or-cancel, an order that will have as much filled immediately as possible with the remainder cancelled
+  PO = 3; // Post Only, an order that will be rejected if any part of the order would match and execute immediately
 }
 
 /**
--- a/broker-cli/proto/broker.proto
+++ b/broker-cli/proto/broker.proto
@@ -309,10 +309,8 @@ service AdminService {
  */
 enum TimeInForce {
   GTC = 0; // Good-til-cancelled, an order that remains in force until cancelled by the user
-  // FOK is not supported
-  // FOK = 1; // Fill-or-kill, an order that can be completely filled or will not be executed
-  // IOC is not supported
-  // IOC = 2; // Immediate-or-cancel, an order that will have as much filled immediately as possible with the remainder cancelled
+  // FOK = 1; // Not supported - Fill-or-kill, an order that can be completely filled or will not be executed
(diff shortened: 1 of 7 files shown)


broker-daemon/block-order-worker/index.js
```diff
@@ -566,19 +566,30 @@ class BlockOrderWorker extends EventEmitter {
...
     const { orders, depth: availableDepth } = await orderbook.getBestOrders({ side: blockOrder.inverseSide, depth: targetDepth.toString(), quantumPrice: blockOrder.quantumPrice })
-    await this._fillOrders(blockOrder, orders, targetDepth.toString())
 
-    if (targetDepth.gt(availableDepth)) {
-      // place an order for the remaining depth that we could not fill
-      this._placeOrders(blockOrder, targetDepth.minus(availableDepth).toString())
+    if (blockOrder.timeInForce === BlockOrder.TIME_RESTRICTIONS.GTC) {
+      await this._fillOrders(blockOrder, orders, targetDepth.toString())
+
+      if (targetDepth.gt(availableDepth)) {
```
treygriffith: can't you do this condition for the post only orders as well?

broker-daemon/block-order-worker/index.js
```diff
@@ -566,19 +566,30 @@ class BlockOrderWorker extends EventEmitter {
...
+        // place an order for the remaining depth that we could not fill
+        this._placeOrders(blockOrder, targetDepth.minus(availableDepth).toString())
+      }
+    } else {
```
treygriffith: might be worthwhile to have an explicit condition here just for clarity of what this branch is doing

broker-daemon/block-order-worker/index.js
```diff
@@ -566,19 +566,30 @@ class BlockOrderWorker extends EventEmitter {
...
+      // Post Only orders are maker only, so if any depth exists for our given
+      // price, we move the order to a cancelled state
+      if (Big(availableDepth).gt(0)) {
+        return this.cancelBlockOrder(blockOrder.id)
```
treygriffith: Add a log here so we know why we're cancelling.

bradleystachurski:
[create post-only order](https://trello.com/c/JOznQj1R/146-create-post-only-order)

broker-cli/proto/broker.proto
```diff
@@ -309,8 +309,11 @@ service AdminService {
  */
 enum TimeInForce {
   GTC = 0; // Good-til-cancelled, an order that remains in force until cancelled by the user
+  // FOK is not supported
```
treygriffith: maybe best to comment out the actual definition of `FOK`

broker-daemon/block-order-worker/index.js
```diff
@@ -566,15 +566,27 @@ class BlockOrderWorker extends EventEmitter {
    * @returns {void}
    */
   async workLimitBlockOrder (blockOrder, targetDepth) {
-    if (blockOrder.timeInForce !== BlockOrder.TIME_RESTRICTIONS.GTC) {
-      throw new Error('Only Good-til-cancelled limit orders are currently supported.')
+    if ((blockOrder.timeInForce !== BlockOrder.TIME_RESTRICTIONS.GTC) && (blockOrder.timeInForce !== BlockOrder.TIME_RESTRICTIONS.PO)) {
+      throw new Error('Only Good-til-cancelled and Post Only limit orders are currently supported.')
```
treygriffith: do you still need this with the other error handler? Seems like it can be removed.

broker-cli/proto/broker.proto
bradleystachurski: @dannypaz Would you be okay with commenting it out?

@treygriffith, Danny and I had an in-person chat about commenting out the definitions in the `enum` and decided to handle the unsupported definition within the implementation rather than having commented out code.

broker-daemon/block-order-worker/index.js
bradleystachurski: Agreed, I'll take it out.

broker-cli/proto/broker.proto
dannypaz: It was a tomato/tomato thing. Basically, we've had these uncommented + throw an exception for unsupported, so we just decided to keep them in

broker-cli/proto/broker.proto
```diff
@@ -309,8 +309,11 @@ service AdminService {
  */
 enum TimeInForce {
   GTC = 0; // Good-til-cancelled, an order that remains in force until cancelled by the user
-  FOK = 1; // Fill-or-kill, an order that can be completely filled or will not be executed
-  IOC = 2; // Immediate-or-cancel, an order that will have as much filled immediately as possible with the remainder cancelled
+  // FOK is not supported
```
treygriffith: [nit] would put this not in the comments next to the proto.

You should also check that the documentation generation from this change makes sense.

broker-daemon/broker-rpc/order-service/create-block-order.spec.js
```diff
@@ -13,7 +13,8 @@ describe('createBlockOrder', () => {
   beforeEach(() => {
     CreateBlockOrderResponse = sinon.stub()
     TimeInForce = {
-      GTC: 1
+      GTC: 1,
```
treygriffith: shouldn't this be `0`?

broker-cli/proto/broker.proto
bradleystachurski: Unfortunately this doesn't look well when rendered.

image

Keeping these uncommented and updating the description to say the time restrictions for `FOK` and `IOC` are not currently implemented. This seems the most intuitive for backwards compatibility on indices, but I'm open to other approaches.

broker-daemon/block-order-worker/index.js
```diff
@@ -566,15 +566,23 @@ class BlockOrderWorker extends EventEmitter {
...
-
     const orderbook = this.orderbooks.get(blockOrder.marketName)
 
-    // fill as many orders at our price or better
+    // get available orders for the requested depth and price
     const { orders, depth: availableDepth } = await orderbook.getBestOrders({ side: blockOrder.inverseSide, depth: targetDepth.toString(), quantumPrice: blockOrder.quantumPrice })
-    await this._fillOrders(blockOrder, orders, targetDepth.toString())
+
+    if (blockOrder.timeInForce === BlockOrder.TIME_RESTRICTIONS.GTC) {
+      await this._fillOrders(blockOrder, orders, targetDepth.toString())
```
treygriffith: [nit] might be nice to add some "why" here, similar to https://github.com/sparkswap/broker/pull/528/files#diff-a509eb0ce29028bbf0f9e9c2ecf846b5L575

PR #535, opened 2019-06-07
Index active orders and fills

treygriffith:
## Description
This change improves the performance of block order creation by speeding up the process of calculating outstanding funds.

In particular, it:
- adds an index of only currently active orders and fills
- parallelizes async operations

In so doing, it also cleans up a lot of the active order/fills calculations.

The performance increase for large datasets (\~150k records) is 100x (\~10s ➡️\~100ms) and should change the growth from linear with total orders and fills to linear with active orders and fills (which is unlikely to grow significantly).

## Related PRs
This is a follow-up to https://github.com/sparkswap/broker/pull/527

## Todos
- [x] Tests

### DIFF
--- a/broker-daemon/block-order-worker/index.js
+++ b/broker-daemon/block-order-worker/index.js
@@ -1,6 +1,8 @@
 const EventEmitter = require('events')
 const { promisify } = require('util')
 
+const ActiveOrdersIndex = require('./active-orders-index')
+const ActiveFillsIndex = require('./active-fills-index')
 const { BlockOrder, Order, Fill } = require('../models')
 const { OrderStateMachine, FillStateMachine } = require('../state-machines')
 const {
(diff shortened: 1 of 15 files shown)


broker-daemon/block-order-worker/index.js
```diff
@@ -61,6 +63,10 @@ class BlockOrderWorker extends EventEmitter {
     const filterOrdersWithOrderId = (key, value) => !!Order.fromStorage(key, value).orderId
     const getOrderIdFromOrder = (key, value) => Order.fromStorage(key, value).orderId
 
+    // create a subset of active orders and fills
```
dannypaz: [nit] maybe to be more clear it could say `create a subset containing active orders and fills`

broker-daemon/orderbook/orderbook-index.js
```diff
@@ -1,39 +1,29 @@
...
-
   /**
    * Create an object that can be passed to Sublevel to create or remove an orderbook record
+   * @private
    * @param {string} key   - Key of the record to create an index op for
    * @param {string} value - Value of the record being added to the events store to create an index op for
    * @returns {Object} object for create/delete for use with sublevel
    */
-  _addToIndexOperation (key, value) {
+  addToIndexOperation (key, value) {
```
dannypaz: Did this API change, or we were just using the internal `_addToIndex..` method instead of the public one?

treygriffith: Just changing the style of private methods to be underscore-less.

PR #539, opened 2019-06-12
Log when Relayer is unavailable during place order

clark800:
## Description
Improved error message for when the broker is trying to place an order and the relayer is unavailable.

## Todos
- [X] Tests


treygriffith:
First PR 🍾 congrats @clark800 !

PR #540, opened 2019-06-13
[wip] Break up orders and fills to fit into available channels

treygriffith:
## Description
This change breaks up orders when being placed, and fills to fit into available channels.

It does in a naive way: simply finding the maximum channel size (or maximum payment size) and making sure every order and fill is smaller than that.

It has a few problems, in particular: one would expect that when repeating this operation multiple times, the second (or nth) operation is destined to fail during settlement. We'll handle this failure in a later change.

It also has a particular egregious issue on fills: it will fill the best priced order only up until its channel limit, and then move to the next best priced order. This could result in the average fill price being worse than expected. Fixing this would require a much more sophisticated approach that will be handled down the road.

## Todos
- [x] Tests

## Future considerations
- Handling failures from capacity at settlement time
- Assigning fills and orders to specific channels to ensure capacity availability
- Filling worst priced orders due to capacity constraints

### DIFF
--- a/broker-daemon/utils/min-big.js
+++ b/broker-daemon/utils/min-big.js
@@ -1,0 +1,17 @@
+const Big = require('./big')
+
+/**
+ * Find the minimum value of a group of Big.js numbers
+ * @param   {...Big} bigs - List of Big.js numbers to compare
+ * @returns {string}        Smallest Big.js number in the set
+ */
+function minBig (...bigs) {
+  return bigs.reduce((min, big) => {
+    if (Big(min).gt(big)) {
+      return big
+    }
+    return min
+  }).toString()
+}
+
+module.exports = minBig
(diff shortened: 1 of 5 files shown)


broker-daemon/utils/min-big.js
```diff
@@ -0,0 +1,17 @@
...
+ * @param   {...Big} bigs - List of Big.js numbers to compare
+ * @returns {string}        Smallest Big.js number in the set
+ */
+function minBig (...bigs) {
+  return bigs.reduce((min, big) => {
+    if (Big(min).gt(big)) {
+      return big
+    }
+    return min
+  }).toString()
```
clark800: Could be more generic to let the caller decide if `toString` is desired

treygriffith:
Closing as out of date.

PR #542, opened 2019-06-14
Add a flag to check the number of records

treygriffith:
## Description
This change adds a flag to the `healthcheck` to count the number of records in every store in the BrokerDaemon. Since this is considered a dev feature, it is only accessible when using the `--json` flag as well.

This should allow more visibility into the Broker and help debug issues related to large data stores.

## Todos
- [x] Tests
- [x] Documentation

### DIFF
--- a/broker-daemon/proto/broker.proto
+++ b/broker-daemon/proto/broker.proto
@@ -141,7 +141,11 @@ message RecordCount {
   int32 count = 3;
 }
 
+/**
+ * HealthCheckRequest is a request about the status and health of a BrokerDaemon.
+ */
 message HealthCheckRequest {
+  // Whether to calculate and include the number of records in the stores of the BrokerDaemon.
   bool include_record_counts = 1;
 }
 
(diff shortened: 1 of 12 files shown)


broker-daemon/proto/broker.proto
```diff
@@ -222,7 +244,7 @@ service AdminService {
    * └───────────────────┴───────────────────┘
    * ```
    */
-  rpc HealthCheck (google.protobuf.Empty) returns (HealthCheckResponse) {
+  rpc HealthCheck (HealthCheckRequest) returns (HealthCheckResponse) {
```
dannypaz: @treygriffith did we want to keep the flag hidden or provide an example in the proto?

dannypaz:
[Need more visibility into the size of anything DB on the broker](https://trello.com/c/iY2LvZZv/1571-need-more-visibility-into-the-size-of-anything-db-on-the-broker)

broker-daemon/proto/broker.proto
treygriffith: No, I think adding it to the docs would be more confusing - I'm fine with it being a hidden field for developer usage.

PR #544, opened 2019-06-18
Allows local Linux setup

zallarak:
## Description
Using `host.docker.internal` to access the host machine does not work on Linux. This naively fixes it. Code changes must also be made to `sparkswap/lnd-engine`.

### DIFF
--- a/scripts/start-sparkswapd.sh
+++ b/scripts/start-sparkswapd.sh
@@ -7,7 +7,16 @@
 #            as `/secure` is a volume inside of the container
 ################################################
 
-set -eu
+set -u
+
+HOST_DOMAIN="host.docker.internal"
+echo $HOST_DOMAIN
+ping -c1 $HOST_DOMAIN > /dev/null 2>&1
+if [ $? -ne 0 ]; then
+  echo "inside if"
+  HOST_IP=$(ip route | awk 'NR==1 {print $3}')
+  echo -e "$HOST_IP\t$HOST_DOMAIN" >> /etc/hosts
+fi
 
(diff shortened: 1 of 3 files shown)


scripts/start-sparkswapd.sh
```diff
@@ -7,7 +7,16 @@
 #            as `/secure` is a volume inside of the container
 ################################################
 
-set -eu
+set -u
+
+HOST_DOMAIN="host.docker.internal"
+echo $HOST_DOMAIN
```
dannypaz: we can remove this echo and the `inside if` below. Also would make a note of why we need to include this inside of the file (because host.docker.internal is used for development networking and `host.docker.internal` doesn't work on linux)

zallarak:
Comments addressed. Pending: testing on Mac.

dannypaz:
[Linux Local Regtest Setup](https://trello.com/c/DjKFyt2R/1705-linux-local-regtest-setup)

scripts/start-sparkswapd.sh
```diff
@@ -7,7 +7,17 @@
 #            as `/secure` is a volume inside of the container
 ################################################
 
-set -eu
+set -u
+
+# Using host.docker.internal to access the host IP does not work for Linux
+# This is a known issue with Docker: https://github.com/docker/for-linux/issues/264
+# Here, we manually map host.docker.internal to the host IP in /etc/hosts
```
dannypaz: One more comment so this description is clearer. We should mention that we map `host.docker.internal` ONLY if the container cannot ping the address (which we assume means we are on linux)

PR #547, opened 2019-06-19
Update swap mechanism for Hold Invoice

treygriffith:
## Description
This change introduces a new Interchain module to replace the Interchain Router and interact with the Hold Invoice functionality added to LND.

It exposes two methods:
- `prepareSwap` which prepares the engines for an upcoming swap
- `forwardSwap` which translates an inbound payment on one chain to another chain to complete its half of the swap

## Todos
- [x] Tests
- [x] Add `cancelSwap` method to LND engine
- [x] Add `settleSwap` method to LND Engine
- [x] Add `waitForSwapCommitment` method to LND engine
- [x] Update `translateSwap` method to use maximum time lock
- [x] Update `prepareSwap` to take an expiration and minimum time lock
- [x] Change `getSettledSwapPreimage` to not wait for invoice changes
- [x] Use in the Order State Machine
- [x] Remove Interchain Router
- [x] Update version of LND Engine


treygriffith:
@clark800  I know what you're saying.

It's two methods because we need to respond back to the Relayer when the swap is prepared, so we need some way to get in the middle of that process. We could switch it to an EventEmitter or similar, but it seems cleaner as two separate methods.

broker-daemon/interchain-bridge/index.js
```diff
@@ -0,0 +1,304 @@
...
+ * Milliseconds between retries of cancels in cases where our timeout has
+ * expired.
+ * @type {number}
+ */
+const RETRY_CANCEL = 10000
+
+/**
+ * Translate a single swap between chains via an Interchain Bridge
+ */
+class InterchainBridge {
```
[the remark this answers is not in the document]
treygriffith: That might be a good idea, let me mull it over.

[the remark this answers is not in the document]
treygriffith: I think the reason I don't like `Swap` is that this is one half of a swap, not the whole thing.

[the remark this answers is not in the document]
clark800: `HalfSwap`?

broker-daemon/interchain/index.js
```diff
@@ -0,0 +1,275 @@
+const { logger } = require('../utils')
```
clark800: Is it still beneficial to have `interchain` be it's own top-level module? Is it only going to be used by the order state machine? If so, we maybe could make this a file in the state-machines module. As a side note, I think conventionally `index.js` usually just presents exports for the public interface of the module.

broker-daemon/interchain/index.js
```diff
@@ -0,0 +1,275 @@
...
+  { engine: outboundEngine, amount: outboundAmount, address: outboundAddress }
+) {
+  // We don't want to reject an incoming HTLC if we know that there is an
+  // active outgoing one. If the outgoing HTLC is in flight or completed, we
+  // should attempt to retrieve the associated preimage.
+  logger.debug(`Checking outbound HTLC status for swap ${hash}`)
+  if (await outboundEngine.isPaymentPendingOrComplete(hash)) {
+    // TODO: return permanent errors when encountered
+    return {
+      paymentPreimage: await outboundEngine.getPaymentPreimage(hash)
```
clark800: Should we just return the preimage rather than putting it in an object? That would make it more consistent with the function name.

broker-daemon/interchain/index.js
```diff
@@ -0,0 +1,275 @@
...
+      throw e
+    }
+
+    logger.error('Temporary Error encountered while translating swap',
+      { error: e.stack, hash })
+
+    // A temporary error means we don't know the current state, so we need
+    // to restart the whole process
+    logger.debug(`Retrying swap translation for ${hash}`)
+    return translateSwap(hash, inboundPayment, outboundPayment)
```
dannypaz: should there be a delay if we continue to retry? I'd assume that there is the potential to get stuck in a loop here

broker-daemon/interchain/index.js
```diff
@@ -0,0 +1,275 @@
...
+ *
+ * @private
+ * @param {string}  hash                  - Base64 string of the hash for the
+ *                                          swap
+ * @param {Payment} inboundPayment        - Expected inbound payment
+ * @param {Engine}  inboundPayment.engine - Engine of the expected inbound
+ *                                          payment
+ * @param {Payment} outboundPayment       - Outbound payment we will make to
+ *                                          retrieve the preimage.
+ * @returns {string}                        Base64 encoded preimage for the swap
```
dannypaz: would add that this also throws an error (jsdoc)

broker-daemon/interchain/index.js
```diff
@@ -0,0 +1,275 @@
...
+    permanentError
+  } = await getPreimage(hash, inboundPayment, outboundPayment)
+
+  if (permanentError) {
+    logger.error(permanentError)
+    logger.error('Downstream payment encountered an error, cancelling ' +
+      `upstream invoice for ${hash}`)
+    await inboundPayment.engine.cancelSwap(hash)
+
+    const err = new Error(permanentError)
```
dannypaz: We have custom errors for other types and although there is no duplication with this specific error, would it be worth it to create `PermanentError` class?

broker-daemon/interchain/index.js
treygriffith: This makes it consistent with the other return values. If we don't do it like this, we have to handle the `permanentError` case inside this function, which I don't think is desirable.

broker-daemon/interchain/index.js
treygriffith: Yes there is. Simple delay? Back-off? What do you think?

broker-daemon/interchain/index.js
treygriffith: That could be a good idea.

broker-daemon/interchain/index.js
clark800: I think you could do it without affecting the error handling, something like this:
`const {preimage} = outboundEngine.translateSwap(...)`
`return preimage`

treygriffith: You have to handle the `permanentError` case, like we do here: https://github.com/sparkswap/broker/pull/547/files#diff-cc53b1d80fb0169112cec8eb19c4fa7aR214

clark800: Oh, I see, `permanentError` is an object property, I was thinking it was an exception. So why isn't it desirable to convert to an exception in this function? It seems cleaner that way from what I can see.

treygriffith: I expect that it will happen in two places. Here (when this is replaced with `TrackPayment`) `outboundEngine.translateSwap`.

clark800: Ah ok, could do a helper function to remove the duplication, but not a big deal either way

treygriffith:
This PR is done for now, but won't get merged until the corresponding changes to LND Engine are made to account for potential changes in its API from what is proposed here.

broker-daemon/state-machines/interchain.js
```diff
@@ -0,0 +1,292 @@
...
+  }
+
+  // wait for an incoming payment to this hash that is accepted,
+  // but not yet settled.
+  try {
+    // TODO: implement this method
+    await inboundEngine.waitForSwapCommitment(hash)
+  } catch (e) {
+    // TODO: fix this error message
+    if (e.isSettled) {
```
treygriffith: @clark800 how do I get access to the custom error types exposed in https://github.com/sparkswap/lnd-engine/pull/234 ? I think I need them to replace my current check.

clark800: They are exported here: https://github.com/sparkswap/lnd-engine/blob/master/src/engine-actions/index.js#L96

broker-daemon/state-machines/interchain.js
```diff
@@ -0,0 +1,291 @@
...
+ * @private
+ *
+ * @param {string}  hash                  - Base64 string of the hash for the
+ *                                          swap
+ * @param {Payment} inboundPayment        - Expected inbound payment
+ * @param {Payment} outboundPayment       - Outbound payment we will make to
+ *                                          retrieve the preimage.
+ * @returns {string}                        Base64 encoded preimage for the swap
+ */
+async function translateOnce (hash, inboundPayment, outboundPayment) {
```
clark800: Since `engine.translateSwap` occurs in `getPreimage`, it looks like we're using `translate` to mean `translateSwap` + `settleSwap`, we might want to use a different name like `executeSwap` for `translateSwap` + `settleSwap`. Also on naming: `getPreimage` sounds like it would only be getting an existing preimage rather than creating a new one with `translateSwap`. Perhaps we could rename `getPreimage` to something like `translateSwapWhenReady` and the "getPreimage" aspect is just the idempotence of the "translateSwap" operation.

clark800:
Looks really good, just left a comment about naming if we want to try to improve readability a bit more

broker-daemon/state-machines/interchain.js
```diff
@@ -0,0 +1,310 @@
...
+ *
+ * @private
+ *
+ * @param {string}  hash                    - Base64 string of the hash for the
+ *                                            swap
+ * @param {Payment} inboundPayment
+ * @param {Engine}  inboundPayment.engine   - Engine of the expected inbound
+ *                                            payment
+ * @param {Payment} outboundPayment
+ * @param {Engine}  outboundPayment.engine  - Engine of the outbound
```
clark800: Since we have a `typedef` for `Payment` above, I think we can remove the redundant description here.

broker-daemon/state-machines/interchain.js
```diff
@@ -0,0 +1,310 @@
...
+ * @param {string}  hash                  - Base64 string of the hash for the
+ *                                          swap
+ * @param {Payment} inboundPayment        - Expected inbound payment
+ * @param {Payment} outboundPayment       - Outbound payment we will make to
+ *                                          retrieve the preimage.
+ * @param {Error}   error                 - Error that caused the retry
+ * @returns {string}                        Base64 encoded preimage for the swap
+ */
+async function retryForward (hash, inboundPayment, outboundPayment, error) {
+  logger.error('Temporary Error encountered while translating swap',
```
clark800: Should we update the error messages to match the new function name?

broker-daemon/state-machines/interchain.js
```diff
@@ -0,0 +1,310 @@
...
+ *                                          retrieve the preimage.
+ * @returns {string}                        Base64 encoded preimage for the swap
+ * @throws {Error} If a permanent error is encountered and the swap is cancelled
+ */
+async function forwardSwap (hash, inboundPayment, outboundPayment) {
+  try {
+    const paymentPreimage = await translateIdempotent(hash, inboundPayment, outboundPayment)
+    logger.debug(`Successfully retrieved preimage for swap ${hash}`)
+
+    // settle the upstream payment prior to returning the preimage
```
clark800: [nit] I think the code is probably clear enough to go without a comment here, and same for the `cancelSwap` comment below. (I like the philosophy from "Clean Code": http://www.kyleblaney.com/software-blog/2012/6/29/comments-are-a-failure.html)

clark800:
Done reviewing, looks good!

PR #548, opened 2019-06-20
stop processing orderbook migration in batches

dannypaz:
This branch is based off of 0.7.4-beta

## Description
A few sentences describing the overall goals of the pull request's commits.

## Related PRs

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Link to Trello

### DIFF
--- a/broker-daemon/utils/migrate-store.js
+++ b/broker-daemon/utils/migrate-store.js
@@ -25,7 +25,7 @@ async function migrateStore (sourceStore, targetStore, createDbOperation, batchS
 
     async function flush () {
       try {
-        await previousBatch()
+        await previousBatch
       } catch (e) {
         return reject(e)
       }
@@ -44,9 +44,7 @@ async function migrateStore (sourceStore, targetStore, createDbOperation, batchS
         flush()
 
         // If a partial batch exists, then we should flush the remaining records
-        if (previousBatch) {
-          await previousBatch()
-        }
+        await previousBatch
 
         return resolve()
       } catch (e) {
--- a/broker-daemon/utils/migrate-store.js
+++ b/broker-daemon/utils/migrate-store.js
@@ -1,5 +1,8 @@
+const { promisify } = require('util')
+
 /**
  * Pipe data from one sublevel store into another
+ *
  * @todo Refactor into an iterator pattern using streams2
  * @param  {sublevel} sourceStore       - Sublevel that is the source of the data to pipe
  * @param  {sublevel} targetStore       - Sublevel that is the destination of the data
@@ -11,64 +14,33 @@ async function migrateStore (sourceStore, targetStore, createDbOperation, batchS
   return new Promise((resolve, reject) => {
     const stream = sourceStore.createReadStream()
 
-    // `batchInProgress` allows us to batch save records in sequence and lock the operation
-    // to make sure we are not inserting records out of the order they are received.
-    //
-    // We set this value to automatically resolve on the first use. This value is then
-    // subsequently set when a batch is processed
-    let batchInProgress = async () => {}
-    let batchResolve
-
+    let previousBatch = async () => {}
     let batch = []
-
-    async function flush (done = () => {}) {
-      if (!batch.length) {
-        return process.nextTick(done)
-      }
-
-      await batchInProgress()
-
-      batchInProgress = () => {
-        return new Promise((resolve, reject) => {
-          batchResolve = resolve
-        })
-      }
-
-      try {
-        targetStore.batch(batch, (err) => {
-          if (err) {
-            batchResolve()
-            return reject(err)
-          }
-
-          batchResolve()
-          done()
-        })
-      } catch (e) {
-        batchResolve()
-        throw (e)
-      }
-
-      // clear the batch
-      batch = []
-    }
 
     stream.on('error', reject)
 
-    stream.on('end', () => {
-      flush(resolve)
+    stream.on('end', async () => {
+      await previousBatch()
+      return resolve()
     })
 
-    stream.on('data', ({ key, value }) => {
+    stream.on('data', async ({ key, value }) => {
       const op = createDbOperation(key, value)
 
-      if (op) {
-        batch.push(op)
+      if (!op) return
+
+      batch.push(op)
+
+      if (!(batch.length >= batchSize)) {
+        return
       }
 
-      if (batch.length >= batchSize) {
-        flush()
-      }
+      await previousBatch()
+
+      previousBatch = promisify(targetStore.batch)(batch)
+
+      // Clear the batch
+      batch = []
     })
   })
 }
--- a/broker-daemon/utils/migrate-store.js
+++ b/broker-daemon/utils/migrate-store.js
@@ -41,7 +41,7 @@ async function migrateStore (sourceStore, targetStore, createDbOperation, batchS
     stream.on('end', async () => {
       try {
         // clear the batch
-        flush()
+        await flush()
 
         // If a partial batch exists, then we should flush the remaining records
         await previousBatch
--- a/broker-daemon/utils/migrate-store.js
+++ b/broker-daemon/utils/migrate-store.js
@@ -26,9 +26,11 @@ async function migrateStore (sourceStore, targetStore, createDbOperation, batchS
 
       if (batch.length < batchSize) return
 
-      stream.pause()
-
       try {
+        // Pause the stream so that we process and transfer records to the targetStore
+        // in the order that they were received
+        stream.pause()
+
         await processBatch(batch)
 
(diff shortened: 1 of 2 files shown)


treygriffith:
Is this WIP? I don't think this actually writes to the database.

dannypaz:
@treygriffith We removed flush from being called in a batch, but we call `flush` on the end of the stream which processes the huge `batch` array

broker-daemon/utils/migrate-store.js
```diff
@@ -7,20 +7,47 @@
...
+async function migrateStore (sourceStore, targetStore, createDbOperation, batchSize = 2000) {
   return new Promise((resolve, reject) => {
     const stream = sourceStore.createReadStream()
+
+    // `batchInProgress` allows us to batch save records in sequence and lock the operation
+    // to make sure we are not inserting records out of the order they are received.
+    //
+    // We set this value to automatically resolve on the first use. This value is then
+    // subsequently set when a batch is processed
+    let batchInProgress = async () => {}
```
treygriffith: you don't have to define this as a function, `await undefined` will immediately resolve.

broker-daemon/utils/migrate-store.js
```diff
@@ -7,20 +7,47 @@
...
+      await batchInProgress()
+
+      batchInProgress = () => {
+        return new Promise((resolve, reject) => {
+          batchResolve = resolve
+        })
+      }
+
+      try {
+        targetStore.batch(batch, (err) => {
```
treygriffith: why not wrap `batch` in `promisify` like we do elsewhere? Then you manipulate the promise directly.

broker-daemon/utils/migrate-store.js
dannypaz: @treygriffith you have to call the function, so you'll receive `batchInProgress is not a function` wont you?

broker-daemon/utils/migrate-store.js
```diff
@@ -1,47 +1,65 @@
...
+    })
 
-      targetStore.batch(batch, (err) => {
-        if (err) return reject(err)
-        done()
-      })
+    stream.on('data', async ({ key, value }) => {
+      // The first time we start we should initialize the previousBatch variable
+      if (!previousBatch) {
+        previousBatch = async () => {}
```
treygriffith: I don't think you need to do this.

I think:

```
let myPromise
await myPromise
```

will resolve on the next tick.

broker-daemon/utils/migrate-store.js
```diff
@@ -1,47 +1,65 @@
...
     let batch = []
 
-    function flush (done = () => {}) {
-      if (!batch.length) {
-        return process.nextTick(done)
+    stream.on('error', reject)
+
+    stream.on('end', async () => {
+      try {
+        if (previousBatch) {
```
treygriffith: I don't think you need the conditional here - `await`ing a non-promise should just resolve.

broker-daemon/utils/migrate-store.js
```diff
@@ -1,47 +1,65 @@
...
     const stream = sourceStore.createReadStream()
+
     let batch = []
 
-    function flush (done = () => {}) {
-      if (!batch.length) {
-        return process.nextTick(done)
+    stream.on('error', reject)
+
+    stream.on('end', async () => {
```
treygriffith: I think you still need to flush the last batch in this function in case you have a partial batch (i.e. < 2000 records).

broker-daemon/utils/migrate-store.js
```diff
@@ -1,47 +1,65 @@
...
+      batch.push(op)
 
-    stream.on('data', ({ key, value }) => {
-      const op = createDbOperation(key, value)
+      if (batch.length < batchSize) return
 
-      if (op) {
-        batch.push(op)
+      try {
+        await previousBatch()
```
treygriffith: I don't think you you want to call `previousBatch` - it looks like it is a promise, not a function.

broker-daemon/utils/migrate-store.js
dannypaz: @treygriffith that is not the functionality that im seeing. You have to call the function don't you (I added this in an earlier comment). I've created a small example, does this look correct?

```
async function dan () {
  console.log('ok')
}

async function sam () {
  await dan
}

sam()
```

There will be no log that is generated

treygriffith: that's because `dan` in this example is a function that returns a promise, not a promise.

`previousBatch` by contrast is the promise itself.

```
const writeBatch = promisify(store.batch)
previousBatch = writeBatch(batch)

...
await previousBatch
```

dannypaz: @treygriffith that was it. ty for that

dannypaz:
@treygriffith check the changes. I decided to use `pause` -> `resume` on the stream. tested this on regtest and looks legit

broker-daemon/utils/migrate-store.spec.js
```diff
@@ -1,117 +1,192 @@
...
+      await onData(params)
+      expect(batchStub).to.have.been.calledBefore(resumeStub)
+    })
+
+    context('batch is full', () => {
+      it('awaits the previous batch if the batch size is large', async () => {
+        migrateStore(sourceStore, targetStore, createDbOperation, 5)
+        const onData = onStub.args[0][1]
+        await onData(params)
+        expect(batchStub).to.not.have.been.called()
```
treygriffith: This seems quite repetitive. Can't you just call `onData(params)` 5 times in a row using a loop and then check that `batchStub` has not been called? Not sure you gain much checking after each go.

Also - seems like 2 or 3 would get the point across just as well as 5.

broker-daemon/utils/migrate-store.spec.js
```diff
@@ -1,117 +1,192 @@
...
+  context('on end', () => {
+    let promisifyStub
+    let batchStub
+    let reverts
 
-    expect(stream.on).to.have.been.calledWith('end', sinon.match.func)
-  })
+    beforeEach(() => {
+      batchStub = sinon.stub()
+      promisifyStub = sinon.stub().returns(batchStub)
```
treygriffith: A `withArgs` here might be nice to be more specific about what the promisify is wrapping.

broker-daemon/utils/migrate-store.spec.js
```diff
@@ -1,117 +1,192 @@
...
+  sinon,
+  expect,
+  rewire
+} = require('test/test-helper')
 
-const migrateStore = require('./migrate-store')
+const migrateStore = rewire(path.resolve(__dirname, './migrate-store'))
 
-describe('migrateStore', () => {
+describe.only('migrateStore', () => {
```
treygriffith: hanging only.

treygriffith: You may want to add a test for processing all elements regardless of batch size. Would be nice to have a single simple test of the end-to-end functionality.

dannypaz: noooooooo

dannypaz: added

broker-daemon/utils/migrate-store.spec.js
```diff
@@ -1,117 +1,192 @@
...
+    reverts.forEach(r => r())
   })
 
-  it('creates a readstream from the store', () => {
-    migrateStore(sourceStore, targetStore, createDbOperation)
+  it('returns a promise', () => {
+    expect(migrateStore(sourceStore, targetStore, createDbOperation, batchSize)).to.be.a('promise')
+  })
 
+  it('creates a read string from the source sublevel store', async () => {
```
treygriffith: `stream` not `string`

broker-daemon/utils/migrate-store.spec.js
```diff
@@ -1,117 +1,192 @@
...
+    expect(batchStub).to.have.been.calledWith([batchEntry])
+    expect(pauseStub).to.have.been.calledBefore(batchStub)
+    expect(batchStub).to.have.been.calledBefore(resumeStub)
+  })
 
-    expect(stream.on).to.have.been.calledWith('end', sinon.match.func)
+  context('on error', () => {
+    it('rejects the promise if there was a failure', () => {
+      const promise = migrateStore(sourceStore, targetStore, createDbOperation, batchSize)
+      const error = onStub.args[2][1]
```
treygriffith: [nit] would be better to use the name of the arg rather than call order to get the error handler.

dannypaz:
[Fix orderbook weirdness on brokers](https://trello.com/c/AnmRXIHx/1745-fix-orderbook-weirdness-on-brokers)

PR #549, opened 2019-06-25
Remove unused onError handler

clark800:
## Description
The onError method of the grpc request object could never be called since the only event is `canceled`, so we can remove it.


broker-daemon/orderbook/index.spec.js
```diff
@@ -396,76 +396,6 @@ describe('Orderbook', () => {
         expect(retries).to.be.eql(1)
       })
     })
-
-    describe('watcher errors', () => {
```
dannypaz: really weird file for the test to be in here

PR #559, opened 2019-07-12
Setup typescript and apply typescript-related fixes

clark800:
## Description
Add typescript check to npm test and apply various JSDoc changes and small refactorings to make the check pass.

### DIFF
--- a/broker-cli/utils/error-handler.js
+++ b/broker-cli/utils/error-handler.js
@@ -1,6 +1,14 @@
 require('colors')
 const { status } = require('grpc')
 
+class CodedError extends Error {
+  constructor (message, code, details) {
+    super(message)
+    this.code = code
+    this.details = details
+  }
+}
+
 /**
  * String to match errors for unregistered entity IDs on the Relayer.
  * @constant
@@ -11,8 +19,8 @@ const NOT_REGISTERED_ERROR = 'not registered'
 /**
  * Takes in an error object and throws a friendly error if the broker daemon is down
  *
- * @param {Error} error
- * @returns {string}
+ * @param {CodedError} error
+ * @returns {string | Error}
  */
 function handleError (error) {
   if (error.code === status.UNAVAILABLE) {
@@ -27,7 +35,7 @@ function handleError (error) {
 /**
  * Handling for internal error messages. Adds action item for user if Relayer encounters an unregistered Broker.
  *
- * @param {Error} error
+ * @param {CodedError} error
  * @returns {string}
  */
 function handleInternalError (error) {
@@ -39,4 +47,6 @@ function handleInternalError (error) {
   }
   return message
 }
+
+void CodedError
 module.exports = handleError
--- a/broker-cli/utils/error-handler.js
+++ b/broker-cli/utils/error-handler.js
@@ -1,13 +1,7 @@
 require('colors')
 const { status } = require('grpc')
 
-class CodedError extends Error {
-  constructor (message, code, details) {
-    super(message)
-    this.code = code
-    this.details = details
-  }
-}
+/** @typedef {import('grpc').StatusObject} GrpcStatusObject */
 
 /**
  * String to match errors for unregistered entity IDs on the Relayer.
@@ -19,8 +13,8 @@ const NOT_REGISTERED_ERROR = 'not registered'
 /**
  * Takes in an error object and throws a friendly error if the broker daemon is down
  *
- * @param {CodedError} error
- * @returns {string | Error}
+ * @param {GrpcStatusObject} error
+ * @returns {string | GrpcStatusObject}
  */
 function handleError (error) {
   if (error.code === status.UNAVAILABLE) {
@@ -35,7 +29,7 @@ function handleError (error) {
 /**
  * Handling for internal error messages. Adds action item for user if Relayer encounters an unregistered Broker.
  *
- * @param {CodedError} error
+ * @param {GrpcStatusObject} error
  * @returns {string}
  */
 function handleInternalError (error) {
@@ -48,5 +42,4 @@ function handleInternalError (error) {
   return message
 }
 
-void CodedError
 module.exports = handleError
--- a/broker-daemon/index.js
+++ b/broker-daemon/index.js
@@ -190,19 +190,20 @@ class BrokerDaemon {
   /**
    * Initializes all block orders for the engine.
    * @param {Promise} enginesAreValidated - a promise that resolves when engines are validated
-   * @returns {void}
+   * @returns {Promise<void>}
    */
   async initializeBlockOrder (enginesAreValidated) {
     this.logger.info(`Initializing BlockOrderWorker`)
     await this.blockOrderWorker.initialize(enginesAreValidated)
     this.logger.info('BlockOrderWorker initialized')
+    return undefined
   }
 
   /**
    * Listens to the assigned markets
    *
    * @param {Array<string>} markets
-   * @returns {void} promise that resolves when markets are caught up to the remote
+   * @returns {Promise<void>} promise that resolves when markets are caught up to the remote
    * @throws {Error} If markets include a currency with no currency configuration
    */
   async initializeMarkets (markets) {
@@ -211,13 +212,14 @@ class BrokerDaemon {
       return this.initializeMarket(marketName)
     }))
     this.logger.info(`Caught up to ${markets.length} markets`)
+    return undefined
   }
 
   /**
    * Creates and initializes an orderbook for every market
    *
    * @param {string} marketName
-   * @returns {void} promise that resolves when market is caught up to the remote
+   * @returns {Promise<void>} promise that resolves when market is caught up to the remote
    */
   async initializeMarket (marketName) {
     const symbols = marketName.split('/')
--- a/broker-daemon/index.js
+++ b/broker-daemon/index.js
@@ -209,6 +209,7 @@ class BrokerDaemon {
     this.logger.info(`Initializing BlockOrderWorker`)
     await this.blockOrderWorker.initialize(enginesAreValidated)
     this.logger.info('BlockOrderWorker initialized')
+    return undefined
   }
 
   /**
@@ -224,6 +225,7 @@ class BrokerDaemon {
       return this.initializeMarket(marketName)
     }))
     this.logger.info(`Caught up to ${markets.length} markets`)
+    return undefined
   }
 
   /**
--- a/broker-daemon/index.js
+++ b/broker-daemon/index.js
@@ -11,9 +11,9 @@ const { logger } = require('./utils')
 const CONFIG = require('./config.json')
 
 /** @typedef {import('lnd-engine')} Engine */
-/** @typedef {Object} Logger */
-/** @typedef {Object} GrpcServerStreaming */
-/** @typedef {Object} GrpcMetadata */
+/** @typedef {object} Logger */
+/** @typedef {object} GrpcServerStreaming */
+/** @typedef {object} GrpcMetadata */
 
 /**
  * Default host and port for the BrokerRPCServer to listen on
@@ -45,8 +45,8 @@ const DEFAULT_RELAYER_HOST = 'localhost:28492'
 /**
  * Create an instance of an engine from provided configuration
  * @param {string} symbol         - Symbol that this engine is responsible for
- * @param {Object} engineConfig   - Configuration object for this engine
- * @param {Object} options
+ * @param {object} engineConfig   - Configuration object for this engine
+ * @param {object} options
  * @param {Logger} options.logger - Logger that this engine should use
  * @returns {Engine}
  * @throws {Error} Unknown engine type
@@ -74,7 +74,7 @@ function createEngineFromConfig (symbol, engineConfig, { logger }) {
  */
 class BrokerDaemon {
   /**
-   * @param {Object} opts
+   * @param {object} opts
    * @param {string} opts.network - current blockchain network of the daemon
    * @param {string} opts.privRpcKeyPath
    * @param {string} opts.pubRpcKeyPath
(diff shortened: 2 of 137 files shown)


broker-daemon/orderbook/price-index.js
```diff
@@ -42,18 +45,22 @@ class PriceIndex extends SublevelIndex {
...
    * Placeholder for implementations of the price index
-   * @param {string} quantumPrice - Decimal of the quantumPrice
+   * @param {string} _quantumPrice - Decimal of the quantumPrice
+   * @returns {string}
    * @throws {Error} keyForPrice must be implemented by child class
    */
-  keyForPrice (quantumPrice) {
-    throw new Error('`keyForPrice` must be implemented by child classes.')
+  keyForPrice (_quantumPrice) {
+    if (_quantumPrice || !_quantumPrice) {
```
treygriffith: might be worth adding a comment for this

broker-daemon/orderbook/price-index.js
```diff
@@ -42,18 +45,22 @@ class PriceIndex extends SublevelIndex {
...
+    return ''
   }
 
   /**
    * Create a read stream of orders with prices at least as good as the given one
    * Note: If no quantumPrice is provided, creates a read stream of all records
-   * @param {string} quantumPrice - Decimal of the price
-   * @returns {ReadableStream}       ReadableStream from sublevel-index
+   * @param {?string} quantumPrice - Decimal of the price
+   * @returns {Sublevel}
```
treygriffith: I think this returns a `ReadableStream`, not a `Sublevel`

broker-daemon/relayer/market-watcher.js
```diff
@@ -32,9 +35,9 @@ class MarketWatcher extends EventEmitter {
 
   /**
    * Delete every event in the store and delay further event processing until its complete
-   * @returns {void} resolves when migration is complete
+   * @returns {Promise<void>} resolves when migration is complete
    */
-  migrate () {
+  async migrate () {
```
treygriffith: why is this `async`? it doesn't use `await`

broker-daemon/relayer/market-watcher.js
```diff
@@ -121,6 +124,7 @@ class MarketWatcher extends EventEmitter {
     } else {
       this.logger.debug(`Unknown response type: ${response.type}`)
     }
+    return undefined
```
treygriffith: unnecessary

broker-daemon/relayer/market-watcher.js
```diff
@@ -107,7 +110,7 @@ class MarketWatcher extends EventEmitter {
     this.logger.debug(`response type is ${response.type}`)
 
     if (RESPONSE_TYPES[response.type] === RESPONSE_TYPES.START_OF_EVENTS) {
-      this.migrate()
+      await this.migrate()
```
treygriffith: This should *not* await, but should instead kick off migration. (Although I'm not sure it makes a difference)

broker-daemon/relayer/market-watcher.js
```diff
@@ -136,22 +140,23 @@ class MarketWatcher extends EventEmitter {
...
-   * @returns {void} Resolves when there are no in-progress promises
+   * @returns {Promise<void>} Resolves when there are no in-progress promises
    */
   async delayProcessing () {
     this.logger.debug(`Waiting for promises to resolve before acting on new response`)
     const promises = Array.from(this.finishBeforeProcessing)
     await Promise.all(promises)
     // clean up our finishBeforeProcessing by removing these resolved promises
     promises.forEach(promise => this.finishBeforeProcessing.delete(promise))
+    return undefined
```
treygriffith: unnecessary

broker-daemon/relayer/market-watcher.js
```diff
@@ -172,12 +177,13 @@ class MarketWatcher extends EventEmitter {
       this.logger.error('Saving market event failed, invalidating sync', { orderId })
       this.emit('error', e)
     }
+    return undefined
```
treygriffith: unnecessary

broker-daemon/state-machines/fill-state-machine.js
```diff
@@ -50,62 +54,67 @@ const FillStateMachine = StateMachine.factory({
...
         // this only defines a getter - it will be set by the `fill` setter
         if (!key) {
+          // @ts-ignore
           return this.fill.key || `${UNASSIGNED_PREFIX}${generateId()}`
         }
+        return key
       },
       additionalFields: {
         /**
-         * @type {StateMachinePersistence~FieldAccessor}
```
treygriffith: getting rid of these helpful annotations seems like a step backwards

broker-daemon/state-machines/fill-state-machine.js
```diff
@@ -231,14 +247,15 @@ const FillStateMachine = StateMachine.factory({
       })
 
       this.logger.info(`Created fill ${this.fill.fillId} on the relayer`)
+      return undefined
```
treygriffith: unnecessary

broker-daemon/state-machines/fill-state-machine.js
```diff
@@ -314,23 +332,24 @@ const FillStateMachine = StateMachine.factory({
       }
 
       this.logger.info(`Filled order ${fillId} on the relayer`)
+      return undefined
```
treygriffith: unnecessary

broker-daemon/state-machines/order-state-machine.spec.js
```diff
@@ -268,18 +268,11 @@ describe('OrderStateMachine', () => {
       await osm.create(blockOrderId, params)
 
       expect(Order).to.have.been.calledWith(blockOrderId, sinon.match(params))
-    })
-
-    it('gets the makerAddress for the order', async () => {
-      await osm.create(blockOrderId, params)
-
       expect(getPaymentChannelNetworkAddressStub).to.have.been.calledTwice()
-      expect(osm.order.makerBaseAddress).to.be.equal('bolt:adsfsdf')
```
treygriffith: why are we getting rid of this test?

broker-daemon/state-machines/plugins/abstract.js
```diff
@@ -9,8 +9,8 @@ class StateMachineAbstractPlugin {
   constructor () {} // eslint-disable-line
 
   /**
-   * A copy of StateMachine~plugin, which is not exposed
-   * @param  {StateMachine} instance       - State machine instance to be hooked
```
treygriffith: this seems like a step backward

broker-daemon/state-machines/plugins/abstract.js
```diff
@@ -31,7 +31,7 @@ class StateMachineAbstractPlugin {
    * State machine plugins expose a `configure` method that gets called against the config object when building the factory
    * By default, we use it to apply our custom transitions
    * @see https://github.com/jakesgordon/javascript-state-machine/blob/master/src/config.js#L119
-   * @param {StateMachine~Config} config - State machine configuration object
```
treygriffith: no way to do a typedef based on the external lib, or not worth it?

broker-daemon/state-machines/plugins/persistence.js
```diff
@@ -1,74 +1,76 @@
 const { promisify } = require('util')
 const StateMachinePlugin = require('./abstract')
 
+/* // JSDoc disabled due to Typescript errors
```
treygriffith: 😕

broker-daemon/state-machines/plugins/persistence.js
```diff
@@ -1,74 +1,76 @@
...
-   * @param  {Object<string, StateMachinePersistence~FieldAccessor>} options.additionalFields - List of additional properties on the instance to be persisted
-   * @param  {string}                                                options.storeName        - Name of the property on the state machine where the StateMachinePersistence~Store is located
+   * @param {object} options
+   * @param {StateMachinePersistenceKeyAccessor} options.key - Name of the property on the state machine instance with the unique key, or a function that derives the key from the instance
+   * @param {AdditionalFields} options.additionalFields - List of additional properties on the instance to be persisted
+   * @param {string} [options.storeName] - Name of the property on the state machine where the StateMachinePersistenceStore is located
    */
-  constructor ({ key = 'id', additionalFields = {}, storeName = 'store' } = {}) {
+  constructor ({ key = 'id', additionalFields = {}, storeName = 'store' } =
+  { key: 'id', additionalFields: {}, storeName: 'store' }) {
```
treygriffith: this seems weird

broker-daemon/state-machines/plugins/persistence.js
```diff
@@ -173,19 +174,19 @@ class StateMachinePersistence extends StateMachinePlugin {
...
     return {
       /**
        * Re-hydrate an OrderStateMachine from storage
-       * @param  {Object} initParams    - Params to pass to the state machine constructor (also to the `data` function)
-       * @param  {Object} options
+       * @param  {object} initParams    - Params to pass to the state machine constructor (also to the `data` function)
+       * @param  {object} options
        * @param  {string} options.key   - Stored key
        * @param  {string} options.value - Plain object of the Order State Machine object
-       * @returns {OrderStateMachine}
```
treygriffith: can't use this type?

broker-daemon/utils/logger.js
```diff
@@ -55,10 +55,7 @@ function createLogger () {
       filterSensitive(),
       winston.format.timestamp(),
       winston.format.json()
-    ),
-    json: true,
```
treygriffith: are these parameters not real?

broker-daemon/index.js
```diff
@@ -211,17 +225,19 @@ class BrokerDaemon {
...
    *
    * @param {string} marketName
-   * @returns {void} promise that resolves when market is caught up to the remote
+   * @returns {Promise<void>} promise that resolves when market is caught up to the remote
    */
   async initializeMarket (marketName) {
     const symbols = marketName.split('/')
-    if (!symbols.every(sym => CONFIG.currencies.find(({ symbol }) => symbol === sym.toUpperCase()))) {
+    if (!symbols.every(sym =>
+      CONFIG.currencies.find(({ symbol }) => symbol === sym.toUpperCase()) !== undefined)) {
```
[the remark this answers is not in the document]
clark800: I think `return` is only necessary when you use curly brackets

broker-daemon/models/market-event.js
```diff
@@ -34,7 +43,7 @@ class MarketEvent {
     this.eventType = eventType
     this.sequence = sequence
     this.payload = payload
-    this.sep = this.constructor.sep
+    this.sep = DELIMITER
```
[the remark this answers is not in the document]
clark800: I don't think so. If we use the constructor we would probably have to `ts-ignore` it.

broker-daemon/models/order.js
```diff
@@ -134,7 +143,7 @@ class Order {
    * @returns {string} 64-bit integer represented as a string
    */
   get baseFillAmount () {
-    return this.fillAmount
+    return this.fillAmount || '0'
```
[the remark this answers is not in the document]
clark800: Yeah, I wasn't sure what to do here, any better ideas?

broker-daemon/orderbook/index.js
```diff
@@ -136,12 +139,13 @@ class Orderbook {
...
-   * @param {Object} args
+   * @param {object} args
    * @param {string} args.side - Side of the orderbook to get orders for (i.e. `BID` or `ASK`)
    * @param {string} args.limit - int64 String of the the amount of orders to return.
-   * @returns {Array<MarketEventOrder>} A promise that resolves MarketEventOrders for the limited records
+   * @returns {Promise<Array<MarketEventOrder>>} A promise that resolves MarketEventOrders
+   *                                             for the limited records
    */
-  getOrders ({ side, limit }) {
+  async getOrders ({ side, limit }) {
```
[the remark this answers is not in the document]
clark800: It's returning `getRecords` which is an async function

broker-daemon/orderbook/price-index.js
clark800: If it is `ReadableStream` then typescript says `Property 'on' does not exist on type 'ReadableStream<any>'`. It looks like it might be sublevel has an extended ReadableStream?

broker-daemon/state-machines/order-state-machine.spec.js
clark800: Because `Order` is being stubbed so it never gets set due to a change in the code; before it was being set in `create` and now it's being set in the `Order` constructor, which isn't being run due to the stub.

broker-daemon/state-machines/plugins/persistence.js
clark800: Yeah, should we just make it always require an argument?

broker-daemon/utils/logger.js
clark800: I think it was for an old version of winston

broker-cli/utils/error-handler.js
```diff
@@ -1,6 +1,14 @@
 require('colors')
 const { status } = require('grpc')
 
+class CodedError extends Error {
```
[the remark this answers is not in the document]
clark800: ServiceError didn't allow `details`, I tried StatusObject and that worked

broker-daemon/models/order.js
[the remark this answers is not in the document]
treygriffith: can we not have it return `string` or `undefined`?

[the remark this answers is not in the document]
clark800: That's what I did initially and it propagates to a bunch of places. I'll push it with that approach so you can see. It does pass with not too many changes.

broker-daemon/orderbook/price-index.js
treygriffith: It's a nodejs stream.Readable, not sure how we can represent that.

broker-daemon/state-machines/plugins/persistence.js
treygriffith: yes that's a good idea

broker-daemon/state-machines/plugins/abstract.js
clark800: I didn't see anything exported by the library that would be useful for this

broker-daemon/state-machines/plugins/abstract.js
clark800: I added a typedef for this, not sure if it's right, but it passes

PR #564, opened 2019-07-16
Add createOrder endpoint

clark800:
## Description
Add `createOrder` endpoint

### DIFF
--- a/broker-daemon/broker-rpc/broker-rpc-server.js
+++ b/broker-daemon/broker-rpc/broker-rpc-server.js
@@ -124,7 +124,7 @@ class BrokerRPCServer {
     this.adminService = new AdminService(this.protoPath, { logger, relayer, engines, orderbooks, store, auth: this.auth })
     this.server.addService(this.adminService.definition, this.adminService.implementation)
 
-    this.orderService = new OrderService(this.protoPath, { logger, blockOrderWorker, auth: this.auth })
+    this.orderService = new OrderService(this.protoPath, { logger, relayer, blockOrderWorker, auth: this.auth, store })
     this.server.addService(this.orderService.definition, this.orderService.implementation)
 
     this.orderBookService = new OrderBookService(this.protoPath, { logger, relayer, orderbooks, auth: this.auth })
--- a/broker-daemon/broker-rpc/order-service/create-order.js
+++ b/broker-daemon/broker-rpc/order-service/create-order.js
@@ -1,7 +1,22 @@
 const { promisify } = require('util')
-const { GrpcResponse } = require('../../utils')
+const { GrpcResponse, Big } = require('../../utils')
 
 /** @typedef {import('../broker-rpc-server').GrpcUnaryMethodRequest} GrpcUnaryMethodRequest */
+
+function normalizeValue (value) {
+  return (new Big(value)).toString()
+}
+
+function normalizeAmount (amount) {
+  return Object.assign({}, amount, { value: normalizeValue(amount.value) })
+}
+
+function normalizeOrder (order) {
+  return Object.assign({}, order, {
+    sourceAmount: normalizeAmount(order.sourceAmount),
+    destinationAmount: normalizeAmount(order.destinationAmount)
+  })
+}
 
 /**
  * Creates an order with the relayer
@@ -10,8 +25,11 @@ const { GrpcResponse } = require('../../utils')
  * @returns {Promise<GrpcResponse>}
  */
 async function createOrder ({ params, relayer, store }) {
-  const { orderId } = await relayer.orderService.createOrder(params)
-  await promisify(store.sublevel('orders').put)(orderId, params)
+  const authorization = relayer.identity.authorize()
+  const timestamp = (new Date()).getTime().toString()
+  const { orderId } = await relayer.orderService.createOrder(params, authorization)
+  const order = Object.assign({}, normalizeOrder(params), { timestamp })
+  await promisify(store.sublevel('orders2').put)(orderId, order)
   return new GrpcResponse({ orderId })
 }
 
--- a/broker-daemon/broker-rpc/order-service/create-order.js
+++ b/broker-daemon/broker-rpc/order-service/create-order.js
@@ -1,0 +1,18 @@
+const { promisify } = require('util')
+const { GrpcResponse } = require('../../utils')
+
+/** @typedef {import('../broker-rpc-server').GrpcUnaryMethodRequest} GrpcUnaryMethodRequest */
+
+/**
+ * Creates an order with the relayer
+ *
+ * @param {GrpcUnaryMethodRequest} orderRequest
+ * @returns {Promise<GrpcResponse>}
+ */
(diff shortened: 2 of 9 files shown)


broker-daemon/broker-rpc/broker-rpc-server.js
```diff
@@ -124,7 +124,7 @@ class BrokerRPCServer {
     this.adminService = new AdminService(this.protoPath, { logger, relayer, engines, orderbooks, store, auth: this.auth })
     this.server.addService(this.adminService.definition, this.adminService.implementation)
 
-    this.orderService = new OrderService(this.protoPath, { logger, blockOrderWorker, auth: this.auth })
+    this.orderService = new OrderService(this.protoPath, { logger, relayer, blockOrderWorker, auth: this.auth, store })
```
treygriffith: I think `blockOrderWorker` is unnecessary

broker-daemon/broker-rpc/order-service/create-order.js
```diff
@@ -0,0 +1,18 @@
...
+
+/**
+ * Creates an order with the relayer
+ *
+ * @param {GrpcUnaryMethodRequest} orderRequest
+ * @returns {Promise<GrpcResponse>}
+ */
+async function createOrder ({ params, relayer, store }) {
+  const { orderId } = await relayer.orderService.createOrder(params)
+  await promisify(store.sublevel('orders').put)(orderId, params)
```
treygriffith: I think this sublevel name might have a collision

broker-daemon/broker-rpc/order-service/create-order.js
```diff
@@ -0,0 +1,18 @@
...
+/** @typedef {import('../broker-rpc-server').GrpcUnaryMethodRequest} GrpcUnaryMethodRequest */
+
+/**
+ * Creates an order with the relayer
+ *
+ * @param {GrpcUnaryMethodRequest} orderRequest
+ * @returns {Promise<GrpcResponse>}
+ */
+async function createOrder ({ params, relayer, store }) {
+  const { orderId } = await relayer.orderService.createOrder(params)
```
treygriffith: you need to authorize the request to the relayer

broker-daemon/broker-rpc/order-service/create-order.js
treygriffith: do we need to normalize any data here? And add a timestamp?

broker-daemon/broker-rpc/broker-rpc-server.js
clark800: This service also contains the block order worker endpoints so I think we need it to maintain the original functionality until we do the switch

broker-daemon/relayer/relayer-client.js
```diff
@@ -90,6 +90,10 @@ class RelayerClient {
 
     this.credentials = channelCredentials
 
+    const orderServiceClient = new this.proto.OrderService(this.address,
```
dannypaz: [nit] inconsistent formatting for this file

broker-daemon/proto/broker.proto
```diff
@@ -664,8 +686,10 @@ message GetTradeHistoryResponse {
  * ```
  */
 service OrderService {
+  rpc CreateOrder (CreateOrderRequest) returns (CreateOrderResponse);
```
dannypaz: we need docs for this.

clark800:
I split out the CreateOrder proto definitions into a separate file that is imported; if we want to stick with this approach it looks like we'd also have to change the way we generate proto files in the relayer so that it produces multiple files. I don't think there's a way to import just specific symbols from another proto file.

broker-daemon/broker-rpc/order-service/create-order.js
```diff
@@ -0,0 +1,37 @@
...
+
+function normalizeValue (value) {
+  return (new Big(value)).toString()
+}
+
+function normalizeAmount (amount) {
+  return Object.assign({}, amount, { value: normalizeValue(amount.value) })
+}
+
+function normalizeOrder (order) {
```
dannypaz: these functions need jsdoc too

broker-daemon/relayer/relayer-client.js
```diff
@@ -89,21 +89,33 @@ class RelayerClient {
     }
 
     this.credentials = channelCredentials
+    const options = { interceptors: [grpcDeadlineInterceptor] }
```
dannypaz: +1 on the changes

proto/relayer.proto
```diff
@@ -1,5 +1,7 @@
 syntax = "proto3";
 
+import "proto/create-order.proto";
```
dannypaz: We can discard these changes until we make modifications to the relayer proto.

These changes will get blown away on a local installation. See `https://github.com/sparkswap/broker/blob/master/scripts/build.sh#L182-L199`

dannypaz: TBH probably the way to go about this is to modify that build script since we already check in the proto. What are your thoughts on how to handle this?

broker-daemon/proto/broker.proto
```diff
@@ -664,8 +665,30 @@ message GetTradeHistoryResponse {
...
+   * which returns an order ID and then the order gets saved to the local
+   * database with the order ID as key.
+   *
+   * ```javascript
+   * const { orderId } = await orderService.createOrder({
+   *   sourceAmount: {
+   *     symbol: 'BTC',
+   *     value: '0.01'
+   *   },
+   *   sourceAddress: 'sb1qsakedujs4exdq5dz2cpldcph6hlvf82l8mn8uk',
```
treygriffith: these addresses of the wrong format. They should be payment channel network addresses of the kind returned by `LndEngine#getPaymentChannelNetworkAddress`

proto/relayer.proto
clark800: I think for now we can just leave the proto files as they were since there isn't really that much duplication and the duplication we do have is somewhat legitimate.

broker-daemon/broker-rpc/order-service/create-order.spec.js
```diff
@@ -0,0 +1,50 @@
...
+      },
+      orderService: {
+        createOrder: sinon.stub().resolves({ orderId })
+      }
+    }
+
+    store = { sublevel: name => { return { put } } }
+  })
+
+  it('creates and stores an order', async () => {
```
treygriffith: might want to add a test for the relayer rejecting the order

proto/relayer.proto
```diff
@@ -240,4 +240,30 @@ service AdminService {
...
+
+message CreateOrderRequest {
+  Amount source_amount = 1;
+  string source_address = 2;
+  Amount destination_amount = 3;
+  string destination_address = 4;
+  RoleRestriction role_restriction = 10;
+}
+
+message CreateOrderResponse {
```
treygriffith: should update this once the relayer proto changes get into master

treygriffith (approved):

